PR #1, opened 2021-12-07
Better error for missing Vault config when using credentials provider

gsmet:


gsmet:
CI is broken until we upgrade to 2.6.0.CR1. I tested it locally. Let's merge.

PR #7, opened 2021-12-08
Finalize vault

gsmet:


integration-tests/vault-app/pom.xml
```diff
@@ -89,7 +89,9 @@
                     <systemPropertyVariables>
                         <!-- see TLS availability in native mode https://github.com/quarkusio/quarkus/issues/3797 -->
                         <!-- to activate tls, set vault-test-extension.use-tls=true and install /opt/graalvm -->
+                        <!--
                         <javax.net.ssl.trustStore>${project.basedir}/src/test/resources/vaultTrustStore</javax.net.ssl.trustStore>
+                        -->
```
gsmet, line 94: @vsevel could you have a look at this commit? If we set the trustStore like this, Maven is then unable to download artifacts with `SSLHandshakeException`. I think it used to work in Core because we downloaded everything before hand.

Tests are passing without this so I'll do that for now so that we can release and make progress.

PR #11, opened 2021-12-09
PKI: Return an empty buffer instead of null when Vault returns 204

kdubb:
This only applies to the `VaultInternalPKISecretEngine.getRaw`. Previously when fetching an empty CA Chain or CRL it would thow a NPE, now it returns an empty list.


vsevel:
can you reflect this change of behavior in the javadoc?

kdubb:
There isn't a change of behavior. This just fixes a NullPointerException when the CA Chain  or CRL is empty.

vsevel:
> There isn't a change of behavior. This just fixes a NullPointerException when the CA Chain or CRL is empty.

ok. I thought we were changing the return type from `null` to empty.

vsevel (approved):

kdubb:
@vsevel Merge?

PR #47, opened 2022-07-08
Support reactive & blocking APIs

kdubb:
* Adds reactive versions of all the APIs (aka engines) with the name `Async` added.
* All the manager beans, which previously implemented the sync APIs, now implement the async APIs
* All the sync API interfaces have been made into beans that forward to the async APIs

### DIFF
--- a/docs/modules/ROOT/pages/includes/attributes.adoc
+++ b/docs/modules/ROOT/pages/includes/attributes.adoc
@@ -1,4 +1,4 @@
-:quarkus-version: 2.8.0.Final
+:quarkus-version: 2.10.1.Final
 :quarkus-vault-version: 1.1.0
 :maven-version: 3.8.1+
 
--- a/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
@@ -192,7 +192,8 @@ Possible values are:
 * low: display all secrets.
 * medium: display only usernames and lease ids (ie: passwords and tokens are masked).
 * high: hide lease ids and dynamic credentials username.
---|`low`, `medium`, `high` 
+-- a|
+`low`, `medium`, `high` 
 |`medium`
 
 
--- a/runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
@@ -4,7 +4,6 @@ import static io.quarkus.vault.runtime.LogConfidentialityLevel.LOW;
 import static io.quarkus.vault.runtime.config.VaultAuthenticationType.APPROLE;
 import static io.quarkus.vault.runtime.config.VaultAuthenticationType.KUBERNETES;
 import static io.quarkus.vault.runtime.config.VaultAuthenticationType.USERPASS;
-import static java.util.concurrent.TimeUnit.SECONDS;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
@@ -13,7 +12,6 @@ import java.nio.file.Paths;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
@@ -31,12 +29,12 @@ import io.quarkus.vault.runtime.client.backend.VaultInternalSystemBackend;
 import io.quarkus.vault.runtime.client.dto.auth.AbstractVaultAuthAuth;
 import io.quarkus.vault.runtime.client.dto.auth.VaultAppRoleGenerateNewSecretID;
 import io.quarkus.vault.runtime.client.dto.auth.VaultKubernetesAuthAuth;
-import io.quarkus.vault.runtime.client.dto.auth.VaultRenewSelfAuth;
 import io.quarkus.vault.runtime.client.dto.auth.VaultTokenCreate;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV1;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2;
 import io.quarkus.vault.runtime.config.VaultAuthenticationType;
 import io.quarkus.vault.runtime.config.VaultBootstrapConfig;
+import io.smallrye.mutiny.Uni;
 
 /**
  * Handles authentication. Supports revocation and renewal.
@@ -49,8 +47,7 @@ public class VaultAuthManager {
     public static final String USERPASS_WRAPPING_TOKEN_PASSWORD_KEY = "password";
 
     private AtomicReference<VaultToken> loginCache = new AtomicReference<>(null);
-    private Map<String, String> wrappedCache = new ConcurrentHashMap<>();
-    private Semaphore unwrapSem = new Semaphore(1);
+    private Map<String, Uni<String>> unwrappingCache = new ConcurrentHashMap<>();
     private VaultConfigHolder vaultConfigHolder;
     private VaultInternalSystemBackend vaultInternalSystemBackend;
     private VaultInternalAppRoleAuthMethod vaultInternalAppRoleAuthMethod;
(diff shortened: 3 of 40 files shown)


kdubb:
FYI, I used Mutiny's `await().indefinitely()` to implement the synchronous APIs. Request timeouts are still respected because they are added in the `VaultClient`.

kdubb:
Also, the implementation of `VaultAuthManager.unwrapWrappingTokenOnce` previously used a semaphore to ensure it only happened once. Since semaphores are not compatible with reactive streams I removed it.

The only downside I can see to this is that in rare cases it might unwrap a value twice. Other than the extra roundtrip I cannot see any other downsides to this.

runtime/src/main/java/io/quarkus/vault/VaultKVSecretAsyncEngine.java
```diff
@@ -0,0 +1,50 @@
...
+import io.quarkus.vault.runtime.config.VaultBootstrapConfig;
+import io.smallrye.mutiny.Uni;
+
+/**
+ * This service provides access to the kv secret engine, taking care of authentication,
+ * and token extension or renewal, according to ttl and max-ttl.
+ *
+ * @see VaultBootstrapConfig
+ */
+public interface VaultKVSecretAsyncEngine {
```
cescoffier, line 15: We avoid the term `async`, which means a lot of things, and is often confused with: we offload the work to a worker thread (i.e., most of the async HTTP client). We use the term reactive to avoid confusion.

So, here the name should be `ReactiveVaultKVSecretEngine.`

cescoffier: This applies to all the interfaces.

docs/modules/ROOT/pages/includes/quarkus-vault.adoc
```diff
@@ -192,7 +192,8 @@ Possible values are:
 * low: display all secrets.
 * medium: display only usernames and lease ids (ie: passwords and tokens are masked).
 * high: hide lease ids and dynamic credentials username.
---|`low`, `medium`, `high` 
+-- a|
```
cescoffier, line 195: For my own understanding, why `a`?

docs/modules/ROOT/pages/includes/attributes.adoc
```diff
@@ -1,4 +1,4 @@
-:quarkus-version: 2.8.0.Final
+:quarkus-version: 2.10.1.Final
```
cescoffier, line 1: does it really require 2.10?

cescoffier (changes requested):
I made a few comments.

However, this extension was not intended to be used by the user directly but to provide the config source, which is inherently blocking. What's your use case?

kdubb:
@cescoffier 
> However, this extension was not intended to be used by the user directly but to provide the config source, which is inherently blocking. What's your use case?

I'm not sure this was ever really true. The extension has always provided interfaces to access secret engines in Vault; starting with the `KV` engines. Additionally, support for the `Transit`, `TOTP`, & `PKI` engines has been added. It also provides interfaces for managing Vault secret engines in general and configuring Vault's authentication methods. At this point a config source is the smallest thing it does haha.

docs/modules/ROOT/pages/includes/quarkus-vault.adoc
kdubb: The formatter that runs during the build makes this change. I have no idea why.

docs/modules/ROOT/pages/includes/attributes.adoc
kdubb: This is another change made during the build.

runtime/src/main/java/io/quarkus/vault/VaultKVSecretAsyncEngine.java
kdubb: Async is just so much shorter! haha.

I understand what you mean by the meaning of async being overloaded. I'll rename them but I'd rather keep the reactive near the end just for code completion's sake.

kdubb: done

vsevel:
> @cescoffier
> 
> > However, this extension was not intended to be used by the user directly but to provide the config source, which is inherently blocking. What's your use case?

vault's extension addresses 2 use cases:
- programmatic access
- config source support

cescoffier:
Ah ok, on the programmatic access, then, yes:

1. It should use the managed vertx instance 
2. The reactive api makes sense

kdubb:
> It should use the managed vertx instance

PR #50, which is stacked on this one, addresses this.

runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
```diff
@@ -197,63 +202,39 @@ private String getPassword() {
...
+                                    "this means that the token has already expired " +
+                                    "(if so you can increase the ttl on the wrapping token) or " +
+                                    "has been consumed by somebody else " +
+                                    "(potentially indicating that the wrapping token has been stolen)";
+                            return new VaultException(message, e);
+                        } else {
+                            return e;
+                        }
+                    })
+                    .memoize().indefinitely();
```
vsevel, line 228: this seems to be doing the trick yes. not a new an issue, but we are probably leaking memory here with that cache.

runtime/src/main/java/io/quarkus/vault/runtime/client/VertxVaultClient.java
```diff
@@ -97,155 +99,155 @@ public VaultException get() {
...
     @Override
-    public void close() {
+    public Uni<Void> close() {
         try {
             if (webClient != null) {
                 webClient.close();
             }
-        } finally {
-            vertx.close();
+        } catch (Throwable ignored) {
```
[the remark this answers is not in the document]
kdubb: Do we want exceptions in close reporting outside?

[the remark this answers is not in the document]
kdubb: Well I guess even in my version that can happen. I was planning on ignoring exceptions during close; I would need to wrap and ignore the `vertx.close()` as well though.

I'm fine with changing it back or implementing the "ignore" correctly.

runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
kdubb: Agreed. I will add a ticket to look at using an expiring cache.

runtime/src/main/java/io/quarkus/vault/runtime/client/VertxVaultClient.java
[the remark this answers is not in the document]
kdubb: Alright. I'll revert to previous.

runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
kdubb: Added #51

vsevel: probably with `VaultDynamicCredentialsManager.credentialsCache` as well, although I doubt anybody will hit an OOM because of that one.

runtime/src/main/java/io/quarkus/vault/runtime/client/VertxVaultClient.java
```diff
@@ -97,155 +99,155 @@ public VaultException get() {
...
-            throw new VaultException(e);
-
-        } catch (io.smallrye.mutiny.TimeoutException e) {
-            // happens if we reach the atMost condition - see UniAwait.atMost(Duration)
-            throw new VaultIOException(e);
+    private <T> Uni<T> exec(HttpRequest<Buffer> request, Object body, Class<T> resultClass, int expectedCode) {
+        Uni<HttpResponse<Buffer>> send = body == null ? request.send()
+                : request.sendBuffer(Buffer.buffer(requestBody(body)));
+
+        return send.ifNoItem().after(getRequestTimeout()).failWith(TimeoutException::new)
```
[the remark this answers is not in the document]
kdubb: This is essentially exactly what was done in the previous code; previously it used `await().atMost(getRequestTimeout())`.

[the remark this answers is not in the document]
kdubb: I've look into `WebClient` and there seems to be no direct "read timeout" setting. I believe the previous code was using this to easily implement a specific "read timeout" setting. The new code is a direct replacement for the old code.

runtime/src/main/java/io/quarkus/vault/runtime/VaultAuthManager.java
```diff
@@ -75,116 +72,124 @@ private VaultBootstrapConfig getConfig() {
...
+                    log.debug(
+                            "created new login token: " + vaultToken.getConfidentialInfo(getConfig().logConfidentialityLevel));
+                    return vaultToken;
+                });
     }
 
-    private VaultToken login(VaultAuthenticationType type) {
-        AbstractVaultAuthAuth<?> auth;
+    private Uni<VaultToken> login(VaultAuthenticationType type) {
+        Uni<? extends AbstractVaultAuthAuth<?>> auth$;
```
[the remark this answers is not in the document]
kdubb: Fixed.

runtime/src/main/java/io/quarkus/vault/runtime/client/VertxVaultClient.java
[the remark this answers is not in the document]
kdubb: Fixed

kdubb:
Thanks for your time reviewing this!

PR #55, opened 2022-07-26
Expire wrapping token cache

### LINKED ISSUE #51: Investigate implementing expiration on infinite caches
The unwrap token cache in `VaultAuthManager` caches unwraps forever. The cache is needed to ensure that each unwrap token is only once.

~Additionally, `VaultDynamicCredentialsManager.credentialsCache` is an infinite cache as well.~

Need to look at expiring the cache entries after a suitably long TTL.

kdubb:
Uses a Caffeine cache with a 1 hour expiry to hold wrapping tokens.

Ostensibly wrapping tokens should be used immediately. The 1 hour window is just to ensure an adequite time window in case of errors or retries.

Fixes #51

### DIFF
--- a/deployment/pom.xml
+++ b/deployment/pom.xml
@@ -43,6 +43,10 @@
             <artifactId>quarkus-smallrye-health-spi</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-caffeine-deployment</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>vault</artifactId>
             <exclusions>
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -41,6 +41,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-caffeine</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-health</artifactId>
             <optional>true</optional>
         </dependency>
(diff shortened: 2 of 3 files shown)


kdubb:
@vsevel Are you ok with the dependency on caffeine to provide this caching ability?

vsevel:
> @vsevel Are you ok with the dependency on caffeine to provide this caching ability?

I will look at this tomorrow. I suppose this is ok on that case. In général I would try avoid depending on extra extensions unless really necessary so we  do not have to force it into the apps 
What do you think @sberyozkin ?

vsevel (approved):

vsevel:
it should not be too hard though for that particular feature to create a minimalist cache. I actually cerated one in `VaultConfigSource` for a different purpose (refresh the properties from the kv secret engine).
it is a matter of replacing the values by a `VaultCacheEntry` and implementing a `clean` in the calling thread (no need to do separate thread to do the cleanup given the size of it).
I think this would be preferable than to force another extension into the app. it is an impl detail. for that trivial ue case, I do not think it is granted.
what do you think?

kdubb:
The caffeine extension is nothing more than an easy way to add the Caffeine library. This is currently a three line change, inventing even a simple expiring cache requires a a lot more code and tests. 

I'd rather not reinvent the wheel even for a trivial use case.

vsevel:
I know. this is more a question about making extensions lean. your opinion @sberyozkin

vsevel:
unless @sberyozkin thinks otherwise, I think we should go ahead with your proposition.
we will listen from the community whether this dependency is seen as damaging. thanks.

kdubb:
Ok. If we are keeping it in the long run replacing any custom cache code with Caffeine would be prudent.

I'll add that as a maintenance issue.

PR #69, opened 2022-08-24
Prepare 2.0.0 Redux

kdubb:


gastaldi (approved):
LGTM! 
Here is a hint that you're not required to follow: I usually set `next-version` to `999-SNAPSHOT`, so you don't need to think about the next development version after a release 😉

kdubb:
Yes, I like the 999-SNAPSHOT thing but this projects has been incrementing the next version.

@vsevel Are you good with using that model?

PR #100, opened 2022-12-22
Handle warning only responses and log warnings

kdubb:
### Warning only responses
When a request expecting a 204 returns warnings it does so by returning a standaard 200 response with the warnings attached. Previously this would result in an exception becaues of the mismatched response status code.

### Logging warnings
Any Vault response can attach warnings. Those warnings are now logged for all responses.

### DIFF
--- a/model/src/main/java/io/quarkus/vault/runtime/client/dto/AbstractVaultDTO.java
+++ b/model/src/main/java/io/quarkus/vault/runtime/client/dto/AbstractVaultDTO.java
@@ -1,5 +1,7 @@
 package io.quarkus.vault.runtime.client.dto;
 
+import java.util.List;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class AbstractVaultDTO<DATA, AUTH> implements VaultModel {
@@ -14,7 +16,7 @@ public class AbstractVaultDTO<DATA, AUTH> implements VaultModel {
     public DATA data;
     @JsonProperty("wrap_info")
     public WrapInfo wrapInfo;
-    public Object warnings;
+    public List<String> warnings;
     public AUTH auth;
 
 }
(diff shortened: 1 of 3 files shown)


kdubb:
@vsevel We need this ASAP, but we need it back on 2.15. How do we go about back point this after merge?

kdubb:
FYI, the reason this is an issue is because we are trying to upgrade to Vault 1.12 and it generates warnings on one of the PKI calls (which we aren't prepared to fix at this point) which turns into an exception because of the 204/200 mismatch.

kdubb:
I pushed a `2.1.x` branch to release a back port from.

vsevel:
> which we aren't prepared to fix at this point

you mean the situation in your infrastructure, which causes the warning?

---

in this code:
```
    @Override
    public String toString() {
        return "VaultClientException{" +
                "operationName='" + operationName + '\'' +
                ", requestPath='" + requestPath + '\'' +
                ", status=" + status +
                ", body='" + body + '\'' +
                '}';
    }
```

may the body contain confidential information? if not we do not want to include it in a `toString` - I see that even in the previous code it was already part of the `toString` :(

---

so here we are not sure what we are going to get?

```
var result = response.bodyAsJson(AbstractVaultDTO.class);
...
        if (!(result instanceof AbstractVaultDTO<?, ?>)) {
            return;
        }
```

what does happen if json is not a `AbstractVaultDTO`? it builds another object (I am surprised it would)? or it fails?
another defensive way to code it, is to use `bodyAsJsonObject()`

---

> I pushed a 2.1.x branch to release a back port from.

ok

kdubb:
> you mean the situation in your infrastructure, which causes the warning?

Yes when using 1.12 warnings are being generated that are not affecting us. I am not sure exactly what they are at the moment.

> may the body contain confidential information?

No. As far as I've ever seen, Vault warnings/errors are not be leaking confidential information. This would be a pretty serious Vault issue and we need the information to be able to be logged.

> so here we are not sure what we are going to get?

Correct but the method `logResultWarnings` is generalized to be called from two cases. The first (you've highlighted) always returns an `AbstractorVaultDTO<>`, the second case _may_ be an `AbstractVaultDTO<>`.

> what does happen if json is not a AbstractVaultDTO?

Nothing. This is to log warnings which we can only get from an `AbstractVaultDTO<>`, if it's not one we don't log anything.

Issue #78, opened 2022-09-15
Vault DB credentials provider with reactive-pg-client


Michal-Kucera:
Hello!

I'm struggling with setting up Vault credentials provider for postgres DB, while using reactive-stack as much as possible. I'm getting the following stack trace:

```
2022-09-15T14:25:21.489236982Z __  ____  __  _____   ___  __ ____  ______ 
2022-09-15T14:25:21.489354273Z  --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
2022-09-15T14:25:21.489362656Z  -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
2022-09-15T14:25:21.489366226Z --\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2022-09-15T14:25:21.492381899Z 2022-09-15 14:25:20,766 WARN  [io.qua.vau.run.TimeLimitedBase] (vert.x-eventloop-thread-1) auth lease duration 60s is smaller than the renew grace period 3600s
2022-09-15T14:25:21.613495011Z 2022-09-15 14:25:21,612 INFO  [io.qua.sma.ope.run.OpenApiRecorder] (main) Default CORS properties will be used, please use 'quarkus.http.cors' properties instead
2022-09-15T14:25:22.144796294Z 2022-09-15 14:25:22,130 ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): javax.enterprise.inject.CreationException: Synthetic bean instance for io.vertx.core.Vertx not initialized yet: io_vertx_core_Vertx_97d170e1550eee4afc0af065b78cda302a97674c
2022-09-15T14:25:22.145156994Z 	- a synthetic bean initialized during RUNTIME_INIT must not be accessed during STATIC_INIT
2022-09-15T14:25:22.145225177Z 	- RUNTIME_INIT build steps that require access to synthetic beans initialized during RUNTIME_INIT should consume the SyntheticBeansRuntimeInitBuildItem
2022-09-15T14:25:22.145243157Z 	at io.vertx.core.Vertx_7c31d5064e6d5db248fafa10f188ca6323f3f983_Synthetic_Bean.create(Unknown Source)
2022-09-15T14:25:22.145272470Z 	at io.vertx.core.Vertx_7c31d5064e6d5db248fafa10f188ca6323f3f983_Synthetic_Bean.create(Unknown Source)
[...]

PR #109, opened 2023-02-06
Avoid obtaining Vertx from CDI

geoand:
This is needed because the VertxVaultClient can potentially be used inside a Synthetic bean and in such cases
obtaining Vertx which is itself a synthetic bean is not possible

Fixes: #78

### DIFF
--- a/runtime/src/main/java/io/quarkus/vault/runtime/client/SharedVertxVaultClient.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/client/SharedVertxVaultClient.java
@@ -2,19 +2,19 @@ package io.quarkus.vault.runtime.client;
 
 import static io.quarkus.vault.runtime.client.MutinyVertxClientFactory.createHttpClient;
 
-import java.lang.annotation.Annotation;
 import java.util.concurrent.atomic.AtomicReference;
 
 import jakarta.annotation.PreDestroy;
 import jakarta.enterprise.context.Dependent;
-import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Singleton;
 
 import io.quarkus.arc.Arc;
+import io.quarkus.arc.ArcContainer;
 import io.quarkus.runtime.TlsConfig;
 import io.quarkus.vault.VaultException;
 import io.quarkus.vault.runtime.VaultConfigHolder;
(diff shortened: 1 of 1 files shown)


kdubb:
@geoand I'm not sure this does exactly what the previous code was doing. I need to look at this more in depth and report back.

The goal with Shared/Private is to ensure the system provided VertX (aka Shared) is used whenever it is available but for config related queries we _might_ need to create a temporary (aka Private) VertX instance to query Vault before the system provides one.

kdubb:
I'm realizing that `createSharedVaultClient` is not the best name for this function. It should at least be `createSharedVaultClientIfAvailable`.

geoand:
Right, that's also what my PR does, doesn't it?

kdubb:
This is a `@Dependent` producer. Previously it was not always creating a new client but only used as a way to lookup the best available client (shared or private). Now it is creating a new client for each injection point.

geoand:
Okay, so how would you like to handle it?

kdubb:
I think the following works. Basically the same code just uses `VertxRecorder.getVerx()` as the flag for shared vs. private and then requests it directly in the constructor instead of injecting it.

```java
@Produces
@Dependent
public static VertxVaultClient createSharedVaultClient() {
    Annotation clientType;
    if (VertxRecorder.getVertx() != null) {
        clientType = Shared.Literal.INSTANCE;
    } else {
        clientType = Private.Literal.INSTANCE;
    }
    return Arc.container().select(VertxVaultClient.class, clientType).get();
}
```

```java
public SharedVertxVaultClient(VaultConfigHolder vaultConfigHolder, TlsConfig tlsConfig) {
    super(vaultConfigHolder.getVaultBootstrapConfig().url.orElseThrow(() -> new VaultException("no vault url provided")),
            vaultConfigHolder.getVaultBootstrapConfig().enterprise.namespace,
            vaultConfigHolder.getVaultBootstrapConfig().readTimeout);
    Vertx vertx = Vertx.newInstance(VertxRecorder.getVertx());
    this.webClient.set(createHttpClient(vertx, vaultConfigHolder.getVaultBootstrapConfig(), tlsConfig));
}
```

kdubb:
I tested this locally. I can push the changes to the PR branch.

geoand:
Sounds good to me

PR #140, opened 2023-06-15
Move config to use mappings

radcortez:
Goal is to move to the new Config mappings, since we want to deprecate the old class style of configs. Also moved away from Bootstrap phase which is already deprecated.

There are some adjustments required on Quarkus processor side (like ignoring the `toString` method in interfaces), which I'm going to improve.


vsevel:
hello @radcortez nice improvement
the only drawback I see are disappearance of custom `toString` methods, where we would display sensitive information according to a log confidentiality level.

see https://github.com/quarkiverse/quarkus-vault/blob/main/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultBootstrapConfig.java#L276

in summary 
- we can't have a `toString()` that prints out all fields, 
- we have to display some non sensitive fields
- ideally, we want to display some sensitive field according to a configurable sensitive level

radcortez:
Yes, that is something that I don't know how to handle.

Mappings are able to generate a plain `toString` method, but obviously, they are not aware of a sensitive level. The `toString` method is generated as is.

Unfortunately, we cannot provide a default method of `toString` (to provide our implementation). Providing some sort of hook to influence the generation of `toString` method seems like an overkill to me.

Would it be acceptable to provide the implementation under a different method signature?

vsevel:
> Would it be acceptable to provide the implementation under a different method signature?

yes.

if we do not support sensitivity, it would be worthwile at least to introduce some annotation `@Confidential` that we can put on individual fields so that they get displayed as `token=XXXXXX`

radcortez:
The latest Quarkus version resolved some issues with the documentation generator, so this should be mostly ready now.

Mappings with plain `toString` use the generated `toString` from the implementation. The one with visibilities, I had to add a new method signature to be able to provide a `toString` implementation. It takes the level as the argument to control what is printed.

> if we do not support sensitivity, it would be worthwile at least to introduce some annotation `@Confidential` that we can put on individual fields so that they get displayed as `token=XXXXXX`

One issue we have with current Mappings and the old classes is that even with such annotation, the secret is still available and cached in the Config. One idea is to provide a specific type to represent secrets. This would allow us to query the secret only when required and never cache the value.

vsevel:
I tried logging a config class with your changes:

```
    public Uni<String> getClientToken(VaultClient vaultClient) {
        log.info("XXXXXXXX " + getConfig().authentication());
        return getConfig().authentication().isDirectClientToken() ? getDirectClientToken(vaultClient)
                : login(vaultClient).map(vaultToken -> vaultToken.clientToken);
    }
```

I get:
```
2023-07-23 19:27:52,667 INFO  [io.qua.vau.run.VaultAuthManager] (main) XXXXXXXX io.quarkus.vault.runtime.config.VaultAuthenticationConfig1335487764Impl@5b0e9e0c
```

is this what you mean when saying _Mappings with plain toString use the generated toString from the implementation. _?
the generated `toString` from the implementation comes from `Object`?

> The one with visibilities, I had to add a new method signature to be able to provide a toString implementation. It takes the level as the argument to control what is printed.

how does it work?
is it supposed to be on the interface?

vsevel:
I see it now @radcortez , in `VaultBootstrapConfig`. I thought it was in other config classes too.

however, the semantic has been broken. when we mask with a tolerance, the argument should be the level at which you may show the value for that particular attribute. it will be different from one attribute to another, and it must be hardcoded (or ideally retrieved from the field definition if we had some annotations).

so instead of:
```
clientToken=" + logConfidentialityLevel().maskWithTolerance(authentication().clientToken().orElse(""), level)
```
this should be:
```
clientToken=" + logConfidentialityLevel().maskWithTolerance(authentication().clientToken().orElse(""),LogConfidentialityLevel.LOW)
```

`kubernetesAuthenticationRole, userpassUsername, appRoleRoleId` are `MEDIUM`
`userpassPassword, appRoleSecretId, appRoleSecretIdWrappingToken, clientToken, clientTokenWrappingToken` are `LOW`

and there is no need for an argument in the new `toString` method.
a suggestion is to call it `toSecuredString` or `toConfidentialString`:

```
    default String toSecuredString() {
        return "VaultRuntimeConfig{" +
                "url=" + url() +
                ...
                ", clientToken="
                + logConfidentialityLevel().maskWithTolerance(authentication().clientToken().orElse(""),
                        LogConfidentialityLevel.LOW)
```

`VaultBootstrapConfig` could be renamed `VaultRuntimeConfig` now.

also there are some conflicts with `main`

once we have this in, I would like to cut a release.

runtime/src/main/java/io/quarkus/vault/runtime/config/VaultBootstrapConfig.java
```diff
@@ -1,42 +1,44 @@
...
-    public static final String DEFAULT_SECRET_CONFIG_CACHE_PERIOD = "10M";
-    public static final String KUBERNETES_CACERT = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt";
-    public static final String DEFAULT_CONNECT_TIMEOUT = "5S";
-    public static final String DEFAULT_READ_TIMEOUT = "5S";
-    public static final String DEFAULT_TLS_USE_KUBERNETES_CACERT = "true";
-    public static final String DEFAULT_KUBERNETES_AUTH_MOUNT_PATH = "auth/kubernetes";
-    public static final String DEFAULT_APPROLE_AUTH_MOUNT_PATH = "auth/approle";
+@ConfigMapping(prefix = "quarkus.vault")
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+public interface VaultBootstrapConfig {
```
vsevel, line 28: rename to `VaultRuntimeConfig`

runtime/src/main/java/io/quarkus/vault/runtime/config/VaultBootstrapConfig.java
```diff
@@ -250,66 +249,76 @@ public class VaultBootstrapConfig {
...
+        } else if (authentication().isAppRole()) {
             return APPROLE;
         } else {
             return null;
         }
     }
 
-    @Override
-    public String toString() {
+    default String toString(LogConfidentialityLevel level) {
```
vsevel, line 285: rename to `toSecuredString()` with no argument
hardcode the level for each attribute, according to their own sensitivity

vsevel (changes requested):

radcortez:
Yes, you require a `toString` method declaration on each interface to generate the implementation.

vsevel (approved):

vsevel:
thanks for this cool update @radcortez !

PR #145, opened 2023-07-05
quarkiverse-parent 15

vsevel:
needed to rename field `sealed` because of error generated in `impsort`: `(line 11,col 12) Parse error. Found "sealed", expected one of  "enum" "exports" "module" "open" "opens" "provides" "record" "requires" "strictfp" "to" "transitive" "uses" "with" "yield" <IDENTIFIER>
`

### DIFF
--- a/docs/modules/ROOT/pages/includes/attributes.adoc
+++ b/docs/modules/ROOT/pages/includes/attributes.adoc
@@ -1,4 +1,4 @@
-:quarkus-version: 3.0.3.Final
+:quarkus-version: 3.2.0.Final
 :quarkus-vault-version: 3.0.0
 :maven-version: 3.8.1+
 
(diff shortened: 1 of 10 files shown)


gastaldi (approved):

kdubb:
@gastaldi Any ideas on this?

gastaldi:
Confirmed. It's a bug in the javaparser. Using a newer dependency fixes the problem:
```xml
            <plugin>
                <groupId>net.revelc.code</groupId>
                <artifactId>impsort-maven-plugin</artifactId>
                <dependencies>
                    <dependency>
                        <groupId>com.github.javaparser</groupId>
                        <artifactId>javaparser-core</artifactId>
                        <version>3.25.4</version>
                    </dependency>
                </dependencies>
            </plugin>

```

gastaldi:
Pull-request created: https://github.com/revelc/impsort-maven-plugin/pull/74

kdubb:
Ok. I'm good with it; just seemed like this was a parser issue and not something we should deal with.

Given the long trail from "JavaParser" -> "Vault Extension", seems like a rename might be in order if this update is important.

PR #146, opened 2023-07-06
Add read secret json

vsevel:
Add support for `readSecretJson(String)` that returns a `Map<String, Object>` , as opposed to a `Map<String, Object> in existing `readSecret(String)` .
This allows reading secrets stored as json documents: `vault kv put secret/foo-json @/tmp/secret.json` 

I decided to add new methods as opposed to change the initial signature because:
- the usual use case is to get strings, and it seems relevant to have a dedicated signature
- changing the original signature brings a lot of changes in many places


kdubb (approved):
May be a nitpick but there are a lot of formatting and upgrade changes included in one commit/PR. It expands what is a pretty small change into quite a lot of changed files.

Maybe these should be done separately? A format update and then this.

vsevel:
> May be a nitpick but there are a lot of formatting and upgrade changes included in one commit/PR. It expands what is a pretty small change into quite a lot of changed files.

is this the case?
I rebase and reformatted but could not see any differences.

kdubb:
It removed all the doc formatting changes and OS windows checks.

I think what got me was those were the first files the review showed me and there were quite a few changes in those files.

vsevel:
or so we are all good I assume.
I wanted your opinion on introducing specific _json_ methods, as opposed to make more general the original read methods (e.g. trying to transform `Uni<Map<String, String>> readSecret(String path);` into `Uni<Map<String, Object>> readSecret(String path);`, or `<T> Uni<Map<String, T>> readSecret(String path);`), with all the cascaded changes in the current code base.

in other words, are you comfortable with adding a specific `readSecretJson` signature @kdubb ?

kdubb:
Absolutely. I looked at that during my review and saw what you saw.

kdubb:
Maybe that's a tech debt task we add? "Generalize `readSecret`"

PR #167, opened 2023-09-14
Don't use VertxRecorder to get Vertx instance

geoand:


vsevel:
can't validate the change, but we are going to fetch your branch and try on a project internally. thanks!

geoand:
👌🏼

kdubb:
@geoand Please look at #109 as I referenced in the other thread. This is almost exactly how I originally wrote it and it was switched to using VertxRecorder in #109 (you opened the PR 😏)

vmargerin:
@vsevel Changes validated on my side

geoand:
Oh right... I have no memory of that whatsoever

kdubb:
Maybe something has changed in 3.4 that allows that to work again?

Here is your original problem statement for #78...
> Essentially the problem is that when creating the synthetic bean for the PgPool, Quarkus loads the CredentialsProvider (via CDI) which then attempts (in this case) to load Vertx from CDI. This last load will fail because Vertx is also a synthetic bean.

kdubb:
I thought this use case had test coverage via quarkus. If it doesn't, it should be add to ensure this doesn't regress as we attempt to change this.

geoand:
It probably works again by luck. 

However I do know what is needed to fix it - I might have time to address the deep problem I described in #78 in Quarkus.

geoand:
https://github.com/quarkusio/quarkus/pull/35949 should fix things once and for all.

geoand:
If worse comes to worse, you can always temporarily use `VertxEventBusConsumerRecorder` instead of `VertxRecorder`

geoand:
@vsevel correct on both counts

kdubb:
@vsevel Go ahead with which fix? This PR? I have trouble accepting a PR that fixes the 3.4 build but we know doesn't work with 3.4. This is near and dear to myself as we use this feature we would be breaking.

IIUC It sounds like we can change this PR to use `VertxEventBusConsumerRecorder` and then if/when #35949 is merged backported merged we then just revert this PR and & #109.

gsmet:
Please use `VertxEventBusConsumerRecorder` for now so that we can get a release compatible with 3.4 before I release the Platform tomorrow around noon.

vsevel:
this PR will continue in the context of quarkus `3.5`

vsevel:
@geoand is this the plan to upgrade to `CR1` today and finalize this PR?
do you need some help?

geoand:
Sure yeah, I can update the PR once CR1 is out

PR #178, opened 2023-10-09
Update version of vault image to latest stable official image 1.15.0 

bmscomp:
The main purpose of this pull request is to use the latest stable version of vault image on `hashicorp/vault` instead the already used one `vault`, as mentioned on https://hub.docker.com/_/vault/  that Upcoming in Vault 1.14, we will stop publishing official `Dockerhub` images and publish only our Verified Publisher images. Users of `Docker` images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault` on this pull request, an update of a recent version of `vault` can keep fresh the library and make it fit to the upcoming new changes and prevent the drastic changes when a lot of breaking changes happens on certain versions 

- Update the vault hub image to use `hashicorp/vault`
- Upgrade the vault version to 1.15.0 
- Use `https` instead of `http` for vault instance
- Use `CertificateException` for `testGetCAChain` and fix the broken test related to this chnage
- Make `CRUD_PATH`  variable final

### DIFF
--- a/deployment/src/main/java/io/quarkus/vault/deployment/DevServicesVaultProcessor.java
+++ b/deployment/src/main/java/io/quarkus/vault/deployment/DevServicesVaultProcessor.java
@@ -30,7 +30,7 @@ import io.quarkus.vault.runtime.config.VaultBuildTimeConfig;
 
 public class DevServicesVaultProcessor {
     private static final Logger log = Logger.getLogger(DevServicesVaultProcessor.class);
-    private static final String VAULT_IMAGE = "vault:" + VaultVersions.VAULT_TEST_VERSION;
+    private static final String VAULT_IMAGE = "hashicorp/vault:" + VaultVersions.VAULT_TEST_VERSION;
     private static final String DEV_SERVICE_LABEL = "quarkus-dev-service-vault";
     private static final String DEV_SERVICE_TOKEN = "root";
     private static final int VAULT_EXPOSED_PORT = 8200;
@@ -182,7 +182,7 @@ public class DevServicesVaultProcessor {
         private final Closeable closeable;
 
         public VaultInstance(String host, int port, String clientToken, Closeable closeable) {
-            this("http://" + host + ":" + port, clientToken, closeable);
+            this("https://" + host + ":" + port, clientToken, closeable);
         }
 
         public VaultInstance(String url, String clientToken, Closeable closeable) {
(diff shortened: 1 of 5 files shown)


kdubb:
@bmscomp Where is `-dev-tls` being set for this container?

kdubb:
Let me elaborate. You changed the protocol from `http` to `https` but this requires `-dev-tls` to be passed to the started container but that isn't being passed. Also, `-dev-tls`  requires a minimum of Vault `1.12`, meaning for people pinned to older Vault versions this won't work.

I've checked `testcontainers` which still defaults to [`Vault 1.1.3`](https://github.com/testcontainers/testcontainers-java/blob/f1f79dc4281f906d7b508b45a79dac28585778d5/modules/vault/src/main/java/org/testcontainers/vault/VaultContainer.java#L31) and uses a default [`http` address](https://github.com/testcontainers/testcontainers-java/blob/f1f79dc4281f906d7b508b45a79dac28585778d5/modules/vault/src/main/java/org/testcontainers/vault/VaultContainer.java#L61)

I've also tried running a local 1.15 container using `docker run --cap-add=IPC_LOCK hashicorp/vault` which defaults to using an `http` address.

So I'm left a bit confused at how this is actually working. `DevServicesVaultProcessor` wasn't updated to pass `-dev-tls` and yet, at the very least, this PR passed CI checks.

kdubb:
@vsevel @bmscomp I think we let this slip by. After a bit of a refresher I realized I don't think we ever actually use "dev services" to start vault for the integration tests (or any tests). We use a custom configuration to boot Vault with TLS support but that has nothing to do with this change.

I don't think dev services works at all at the moment. @bmscomp Have you tried actually tried using this in a project with dev services active?

How I discovered this is we are finally upgrade to Q3.5.2 and basically none of our Vault dev services start because the containers are started using `http` and the address was changed to `https`

bmscomp:
I am working on that right mow

vsevel:
> How I discovered this is we are finally upgrade to Q3.5.2 and basically none of our Vault dev services start because the containers are started using http and the address was changed to https

we were not cautious enough on this. thanks @kdubb for raising a flag.

bmscomp:
@kdubb  @vsevel  I think it's a good idea to add an integration or system test to make sure such important feature have no regression, rather then let this for a manual test, I'll try next days to work on something like this

kdubb:
@bmscomp It should be very easy. Any integration test having a configuration that does _not_ provide a vault url/credentials should automatically start devservices.

PR #179, opened 2023-10-10
Remove unnecessary  return statements on lambdas and encapsulate public fields with fluent accessors

bmscomp:
The purpose of this pull request is to bring some practices for reducing the number of lines on a growing class, and apply some practices, with some formatting to make it more readable and less verbose, notice that the maven formatter plugged into the project is used to ensure the conformity of the formatting

- remove non necessary return statement on some lambdas
- replace a check of instance of by non nullable check
- remove duplicated cast  for List<?>
- encapsulated directly accessed public fields with fluent accessors

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
@@ -260,13 +260,13 @@ Renew grace period duration.
 <p>
 This value if used to extend a lease before it expires its ttl, or recreate a new lease before the current
 lease reaches its max_ttl.
-By default Vault leaseDuration is equal to 7 days (ie: 168h or 604800s).
+By default, Vault leaseDuration is equal to 7 days (ie: 168h or 604800s).
 If a connection pool maxLifetime is set, it is reasonable to set the renewGracePeriod to be greater
 than the maxLifetime, so that we are sure we get a chance to renew leases before we reach the ttl.
 In any case you need to make sure there will be attempts to fetch secrets within the renewGracePeriod,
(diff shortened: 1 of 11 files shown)


kdubb:
IMO, many of these changes make the code _less_ readable. There doesn't seem to be any advantage to replacing these single statement blocks with expressions, except to push the code 4-8 spaces to the right.

bmscomp:
@kdubb  I agree the pull request is too minimal and it contains only a remove of the duplication and the nullability check instead of instance of but I can make it rich by updating it with other related topics

bmscomp:
@kdubb  I updated the pull request to add a final keyword to a field in class and of course encapsulate fields `SignIntermediateCAOptions.java` and I used fluent accessors, as java suggest for records, to prevent non used methods

kdubb (commented):
The biggest piece of this PR is source breaking and cannot be implemented without a major version bump. Additionally, the other "options" classes in the API are implemented similarly so we'd be changing just one of them, not all.

Finally, a point of personal preference (seemingly shared by @vsevel), there is no advantage to adding encapsulating fields with public getters/setters for very simple classes like these "options" classes. Yes there can be advantages when the class/object will be long-lived or needs to interface with an API expecting a "bean". These objects are very transient and 99% of the time only live as long as a single function call.

I'm not sure this is something we need to tackle now or in the future.

WRT other changes. There are some documentation fixes that could be moved into another PR, which would be helpful.

kdubb:
This is mostly implemented in #215 and the anyhting else is quite of out-of-date because of that PR.

PR #180, opened 2023-10-10
Class Literal implements annotation interface Private 

bmscomp:
The class Literal implements annotation interface Private, it's better to not implement an Annotation type declaration cannot have explicit superinterfaces, so better to remove it from parent interfaces of Literal

### DIFF
--- a/runtime/src/main/java/io/quarkus/vault/runtime/client/Private.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/client/Private.java
@@ -10,7 +10,7 @@ import jakarta.inject.Qualifier;
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Private {
 
-    class Literal extends AnnotationLiteral<Private> implements Private {
+    class Literal extends AnnotationLiteral<Private> {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/io/quarkus/vault/runtime/client/Private.java
```diff
@@ -10,7 +10,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Private {
 
-    class Literal extends AnnotationLiteral<Private> implements Private {
+    class Literal extends AnnotationLiteral<Private> {
```
gsmet, line 13: I think we want to keep this. This is the pattern we use for ``AnnotationLiteral``s (even if not recommended by IDEs).

gsmet (changes requested):

runtime/src/main/java/io/quarkus/vault/runtime/client/Private.java
kdubb: Agreed. This is how they're implemented; not just in Quarkus.

PR #182, opened 2023-10-11
Quarkus 3.5.0.CR1 + Don't use VertxRecorder to get Vertx instance

vsevel:
supersedes https://github.com/quarkiverse/quarkus-vault/pull/167 @geoand 
relates to https://github.com/quarkusio/quarkus-platform/pull/1002#issuecomment-1757889375 @gsmet 
strangely, I had to rework the assertions on the pki test. what do you think @kdubb ?

### DIFF
--- a/docs/modules/ROOT/pages/includes/attributes.adoc
+++ b/docs/modules/ROOT/pages/includes/attributes.adoc
@@ -1,4 +1,4 @@
-:quarkus-version: 3.4.2
+:quarkus-version: 3.5.0.CR1
 :quarkus-vault-version: 3.2.0
 :maven-version: 3.8.1+
 
--- a/integration-tests/vault/pom.xml
+++ b/integration-tests/vault/pom.xml
@@ -13,6 +13,16 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk18on</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk18on</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit5-internal</artifactId>
             <scope>test</scope>
--- a/integration-tests/vault/src/test/java/io/quarkus/vault/VaultPKIITCase.java
+++ b/integration-tests/vault/src/test/java/io/quarkus/vault/VaultPKIITCase.java
@@ -4,16 +4,16 @@ import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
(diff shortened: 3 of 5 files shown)


kdubb:
The changes to assertions raise an eyebrow. After looking I'm assuming they are due to a change in the BouncyCastle version for 3.5?

kdubb:
This ups bctls... https://github.com/quarkusio/quarkus/pull/36331

vsevel:
same. no idea. does this ring a bell @geoand ?

geoand:
Non whatsoever

kdubb:
@vsevel In your local can you pin `bctls` back and see if that's the cause? Or maybe I'm making too much of it.

kdubb:
It seems to be from `ASN1Encodable.toString` which isn't our concern. Just give me a second to check it.

vsevel:
`certificate.getExtension(subjectAlternativeName).getParsedValue()` prints `[[CONTEXT 0][1.3.6.1.4.1.311.20.2.3, [CONTEXT 0]test], [CONTEXT 2]#616c742e6578616d706c652e636f6d, [CONTEXT 7]#01020304, [CONTEXT 6]#65783a3132333435]`

vsevel:
when I decode the generated certificate with https://certlogik.com/decoder/ I get SAN 
```
            X509v3 Subject Alternative Name: 
                othername:<unsupported>, DNS:alt.example.com, IP Address:1.2.3.4, URI:ex:12345
```

kdubb:
So the issue I'm concerned with is instability in our dependencies. After tracing them and debugging I'm seeing the BouncyCastle come from testcontainers. We should be getting it from Quarkus, no?

vsevel:
yes `ASN1Util` shaded from testcontainers

```
    public static String getTagText(int var0, int var1) {
        switch (var0) {
            case 64:
                return "[APPLICATION " + var1 + "]";
            case 128:
                return "[CONTEXT " + var1 + "]";
```

I had no idea we were using bc to begin with.
obviously, ideally we woud use the same lib than in prod.

kdubb:
Admittedly this is my mistake as I didn't check the provenance of the bouncy castle dependency and we are only using it in the test, given that maybe just go with it.

I'll open an issue to align BC with Quarkus.

kdubb:
@vsevel Nice! Does that auto-align from quarkus via BOM?

vsevel:
I have added this this is coming from `jdk18on-1.76`:
```
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcprov-jdk18on</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcpkix-jdk18on</artifactId>
            <scope>test</scope>
        </dependency>
```

and corrected the imports.
it does not change the _new_ behavior.

kdubb:
> it does not change the new behavior.

Fine as long as we're aligned!

vsevel:
that was me doing the last upgrade back then (9 months ago)!
https://github.com/quarkusio/quarkus/commit/288a8bd1d4841101ed549c53d14436f35130938e

I do not know why we have the change in behavior only now...

vsevel:
ok let's wait for the build. you validate?

kdubb:
👍

PR #184, opened 2023-10-15
Enhance programmatic access to the KV secret engine

bmscomp:
The purpose of this pull request is to enhance the programmatic access to the KV secret engine, to give it the ability to access a multi mount KV secret engine, it comes that in some production environnement secrets are stored on different mounts for different domains, and this is not allowed in the current version of Quarkus vault extension, even when using the programatic way it comes always to retrieve the secrets only the configured mount set up in the properties file 

In the current pull request we overloaded some methods to give the extension this ability


kdubb:
The `PKI` engine uses a different method. It simply allows creating an "engine" for a given "mount"; instead of overloading every conceivable method.

I think that's the model to follow.

bmscomp:
@kdubb  Yes it's a good idea I need a name for this engine, if you can help on finding a name for it it will be great for me, yes I agree it's better to define a new engine rather then overload the existing engine and method, notice that the other engine will always be of kind KV

kdubb:
You don't need a new engine. You just need to allowing creating `VaultKVManager` with a specific mount point (like `PKI` does).

bmscomp:
@kdubb  I can get it now, I'll give it a try

bmscomp:
@kdubb I made some changes for the pull request now the Oliver loaded methods are still here, it may be a kind of shortcut, in the next tastemaker I'll remove it

bmscomp:
@kdubb I updated the pull request and I also rebased it please can you give it a check, Thanks in avance

bmscomp:
@kdubb please provide me with a feedback after the latest changes

kdubb:
@bmscomp Not ignoring you or your contribution. I am swamped with current work and it's going to be a while before I can weigh in on this.

bmscomp:
@kdubb  Thank you su much for the feedback

runtime/src/main/java/io/quarkus/vault/VaultKVSecretEngine.java
```diff
@@ -18,7 +18,8 @@
 @ApplicationScoped
 public class VaultKVSecretEngine {
 
-    private final VaultKVSecretReactiveEngine engine;
+    @Inject
```
kdubb, line 21: Why this change? It was already using constructor injection into a final field?

runtime/src/main/java/io/quarkus/vault/runtime/VaultKvManager.java
```diff
@@ -33,23 +41,48 @@ public static VaultKvManager privateClientManager(@Private VaultClient vaultClie
...
             VaultInternalKvV2SecretEngine vaultInternalKvV2SecretEngine) {
+        this.mount = mount;
         this.vaultClient = vaultClient;
         this.vaultAuthManager = vaultAuthManager;
         this.vaultConfigHolder = vaultConfigHolder;
         this.vaultInternalKvV1SecretEngine = vaultInternalKvV1SecretEngine;
         this.vaultInternalKvV2SecretEngine = vaultInternalKvV2SecretEngine;
     }
 
+    public static VaultKvManager of(String mount,
```
kdubb, line 72: Why add a static constructor? The constructor is only ever called by CDI or the engine factory?

model/src/main/java/io/quarkus/vault/runtime/client/dto/kv/VaultKvSecretV2WriteBody.java
```diff
@@ -2,11 +2,43 @@
 
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import io.quarkus.vault.runtime.client.dto.VaultModel;
 
 public class VaultKvSecretV2WriteBody implements VaultModel {
```
kdubb, line 9: These DTO classes are internal and will most be generated at some point in the future. No need to clean them up or change them at this point.

runtime/src/main/java/io/quarkus/vault/runtime/VaultKvManager.java
```diff
@@ -60,15 +93,16 @@ private Map<String, String> convert(Map<String, Object> map) {
...
 
     @Override
     public Uni<Map<String, Object>> readSecretJson(String path) {
-        return vaultAuthManager.getClientToken(vaultClient).flatMap(token -> {
-
-            String mount = getConfig().kvSecretEngineMountPath();
+        return readValues(mount, path);
+    }
 
+    private Uni<Map<String, Object>> readValues(String mount, String path) {
```
kdubb, line 104: Can't you name this `readSecret` as well? It has a mount path to handle the overload.

runtime/src/main/java/io/quarkus/vault/runtime/VaultKvManager.java
```diff
@@ -77,27 +111,30 @@ public Uni<Map<String, Object>> readSecretJson(String path) {
         });
     }
 
-    @Override
-    public Uni<Void> writeSecret(String path, Map<String, String> secret) {
+    private Uni<Void> writeValues(String mount, String path, Map<String, String> secret) {
```
kdubb, line 114: `writeSecret`? No need to change all the names.

runtime/src/main/java/io/quarkus/vault/runtime/VaultKvManager.java
```diff
@@ -107,16 +144,15 @@ public Uni<Void> deleteSecret(String path) {
         });
     }
 
+    private Uni<List<String>> listValues(String mount, String path) {
```
kdubb, line 147: `listSecrets`?

kdubb (changes requested):
Structure is good and similar to PKI thus providing a consistent method for programmatic access.

Just need to minimize unnecessary changes.

bmscomp:
@kdubb  I'll rebase the brach and solve conflicts and then take care of the comment's reviews by the end of day

kdubb:
@bmscomp Please look at #214 & #215. It does this and a whole lot more.

If you can please build it locally and comment on #214 with ideas and/or #215 with specific suggestions for changes.

bmscomp:
@kdubb Thanks for your answer, I checked the great work done to bring up a new client to vault extension for Quarkus, and I think the content of this pull request  still remains important as an evolution to the programmatic way to support multi without any breaking changes to the existing way to use a programatic way to get vault secrets, we can merge this pull request then after try to enhance the api with your implementation of vault client, since we still need in some cases something strongly coupled with Quarkus, this will also enrich the existing API, Please would you consider the updated review 

Thanks in advance

bmscomp:
@kdubb @vsevel  please any feedback about the current pull request

kdubb:
@bmscomp The new client is still deeply integrated into Quarkus. It has the added ability to be run standalone without using Quarkus static configuration. Additionally, all of the apis that support a dynamic mount path (e.g., any "secret" api like kv) can be easily accessed.

For example, to access a kv engine mounted at a specific path using the new client you would do this:

```java
@Inject VaultClient client; // Inject Quarkus configured client

Map<String, Object> getSomeSecret(String mount) {
   return client.secrets().kv(mount).read("something").getData();
}
```

If you want a "preconfigured" mount you can create a CDI `@Producer` for it.

```java
@Produces
VaultSecretsKV2 getSecretAPI(VaultClient client, @ConfigProperty("app.kv.mount") String mount) {
   return client.secrets.kv(mount);
}
```

And use it like so:

```java
@Inject VaultSecretsKV2 kv2;

Map<String, Object> getSomeSecret() {
  return kv2.read("something").getData();
}
```

Given the advantages of the new client. The current thinking is to deprecate and remove the current "engines" because having and maintaining two separate implementations is a burden. 

Currently PR #215 re-implements the "engines" using the new client simply to allow using it in its current form and validate that the new client is solid and performing correctly.

You are advocating for enhancing the current KV engine at the same time. This would keep users using the "old" system when we want them to be replacing it by using the new client approach.

kdubb:
@bmscomp I understand that this new client PR undermines the work you've done in this PR, but it was all the work that you were required to do that inspired me to get to work on the new client.

Transforming each engine into one that can be both statically configured and dynamically configured, as you now know, is quite a bit of work. Additionally, the CDI model is a bit complicated to ensure both a reactive and non-reactive api exists for _everything_.

I wanted to get away from this model... quickly. It slows down development and introduces a maintenance headed in the long run. The new client solves all these problems and more, all the while making development, whether adding new APIs or maintaining old ones, easier.

bmscomp:
@kdubb  Then would you merge your pull request

kdubb:
Waiting for review. I've been in contact with @vsevelv and he's going to look at it early this week.

bmscomp:
@kdubb  Can you please make public my answers to your reviews, at least other developers or reviews or developers can suggest more

kdubb:
@bmscomp I have no idea what you're requesting. I believe everything in Quarkiverse projects is public (and we team members do not control the project settings, that's done by the Quarkus admins).

bmscomp:
@kdubb  it is totally my mistake apologies, I just could not submit them I think

bmscomp:
Please @gsmet @kdubb  Can you take a look at this pull request, it's updated regarding the last comments and we really need this feature

bmscomp:
Any news about this pull request ?

kdubb:
`4.0.0-alpha.2` has been released. Injecting the `VaultClient` (via `@Inject VaultClient client`) and using `client.secrets().kv1(mountPath)` or `client.secrets().kv2(mountPath)` allows you to supply any mount path.

I'm closing this now that dynamic mounts should use the client directly instead.

Issue #189, opened 2023-10-30
[3.2.0]: Boolean entry in vault no more supported


edelfour:
Hello,
We are currently migrating to Quarkus 3.4.3 which uses quarkus-vault:3.2.0  
In our vault, we have entries of boolean type, for example: "dev-mode":true   
At the startup we encoutered the following exception:
```
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR java.lang.RuntimeException: Failed to start quarkus
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runtime.Application.start(Application.java:101)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:111)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runtime.Quarkus.run(Quarkus.java:71)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.runner.GeneratedMain.main(Unknown Source)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR Caused by: java.lang.ClassCastException: class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at io.quarkus.vault.runtime.VaultKvManager.lambda$convert$0(VaultKvManager.java:58)
   2023-10-28T21:09:54.61+0000 [APP/PROC/WEB/0] ERR at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Unknown Source)
[...]

PR #190, opened 2023-10-30
VaultKvManager: Convert using String.valueOf

edelfour:
in private convert method use String.valueOf instead of String cast to convert value fetched from vault.

resolve quarkiverse/quarkus-vault#189

### DIFF
--- a/integration-tests/vault/src/test/java/io/quarkus/vault/VaultProxyITCase.java
+++ b/integration-tests/vault/src/test/java/io/quarkus/vault/VaultProxyITCase.java
@@ -38,6 +38,8 @@ class VaultProxyITCase {
         assertEquals(PROXY_HOST, host);
         assertEquals(PROXY_PORT, port);
         assertEquals("{hello=world}", kvSecretEngine.readSecret("foo").toString());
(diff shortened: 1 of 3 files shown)


edelfour:
Hello @gsmet 
I fixed the formatting error which failed the pipeline.

gsmet:
@edelfour that being said, the tests are not passing AFAICS

edelfour:
Hello @gsmet 
Thanks, I'm on it. I have run a just `mvn clean test` on my local environment instead of `./mvnw -B formatter:validate verify --file pom.xml -Dnative` in the pipeline configuration.
Sorry for the disturbance

vsevel:
> @kdubb @vsevel any chance one of you could have a look? The change makes sense to me but I would prefer if one of you could check and approve it.

@gsmet yes. we have been discussing in the issue https://github.com/quarkiverse/quarkus-vault/issues/189

edelfour:
a better solution was proposed by @vsevel in https://github.com/quarkiverse/quarkus-vault/issues/189  
I close the PR  
Thanks for your help :-)

PR #194, opened 2023-11-08
Support scalar value other than String from Vault

edelfour:
This pull request modify the way VaultConfigSource fetches the values from Vault following recommendation made by @vsevel 
Now, it fetches the value as Object and convert it to String if type is not List or Map.  
null value is also managed.  
Also add a note in documentation to mention that list and map are not yet supported.

resolve quarkiverse/quarkus-vault#189


vsevel (approved):

edelfour:
thanks :-)
squash and rebase done

vsevel:
thanks @edelfour

PR #198, opened 2023-11-19
Revert to using HTTP when connecting to dev services containers

kdubb:
This fixes dev services by using the correct scheme, HTTP, when connecting to automatically started dev services containers.

@vsevel I'm thinking we revert this to HTTP instead of passing the new `-dev-tls` flag and using HTTPS. My reasoning is just, as we've discussed previously, security products are generally updated slowly and the new flag requires Vault `1.12+`.

We could selectively enable TLS but we'd need to be figuring it out prior to container startup based on the image tag. It can be done it's just work I don't have time for and we need to get this back in working order.

Also, we really need to add a unit test that actually uses a proper "dev services" container. Unfortunately I'm swamped at the moment.

### DIFF
--- a/deployment/src/main/java/io/quarkus/vault/deployment/DevServicesVaultProcessor.java
+++ b/deployment/src/main/java/io/quarkus/vault/deployment/DevServicesVaultProcessor.java
@@ -182,7 +182,7 @@ public class DevServicesVaultProcessor {
         private final Closeable closeable;
 
         public VaultInstance(String host, int port, String clientToken, Closeable closeable) {
-            this("https://" + host + ":" + port, clientToken, closeable);
+            this("http://" + host + ":" + port, clientToken, closeable);
         }
 
         public VaultInstance(String url, String clientToken, Closeable closeable) {


kdubb:
If you're not up-to-date on what this is related to see #178 where I originally commented on the closed PR

bmscomp:
@kdubb  Thanks for the PR I was working right now on a same fix

kdubb:
@bmscomp Can you add a test that actually uses a dev services container so we don't run into this problem again?

kdubb:
We _must_ merge this ASAP to ensure dev services is restored for before Quarkus 3.6.0 is released.

@gsmet If I make a `3.3.1` version of the extension is that something that can be back ported into `3.5.3`? Currently dev services is broken for the 3.5 series.

vsevel:
I was planning a release anyway for `3.6.0` (quarkus vault `3.4` then)
we can add a `3.3.1` for quarkus `3.5`.

vsevel (approved):

kdubb:
@vsevel Should we cut a CR2 just to ensure a "release" works (since we don't yet have a test).

vsevel:
see https://github.com/quarkiverse/quarkus-vault/pull/199

PR #206, opened 2023-12-07
Update sys backend to include 1.15 options & add support for managing plugins

kdubb:
* Updates Secret engine enabling options, info & tune info to include all parameters for Vault 1.15
* Adds support for listing, getting, registering and removing plugins.

Note about plugins...

I added a "custom" plugin to ensure the plugin management works and is tested. To do this it requires adding variants for arm and x86 archs to ensure it runs on _most_ hardware (most importantly CI). The plugin is a stock build from the latest "hashicups" example secret engine. I then used `udx --ultra-brute` to smash them down as small as possible.

The plugin is "unused", it is only there so that we can properly test mounting a secret engine with a custom plugin. Given the timeline of Vault plugin support these binaries should be good for quite a long time (many years); but at some point they will probably have to be recreated.

[Repo for plugin](https://github.com/hashicorp-education/learn-vault-plugin-secrets-hashicups)


vsevel (approved):
LGTM nice addition!

PR #211, opened 2023-12-09
Support all tunable values for `updateTuneInfo`

kdubb:
`updateTuneInfo` now correctly updates all the fields available for tuning. Additionally, the missing plugin version was added to `VaultTuneInfo`.


kdubb:
Implemented in #215

PR #237, opened 2024-02-15
JDK 17

vsevel:


kdubb:
@vsevel Isn't this already done in #238?

vsevel:
> @vsevel Isn't this already done in #238?

yes @kdubb 
this PR was originally intending to move to CR1 then 3.8.0, plus JDK 17
the only thing to do is JDK 17
can you validate?
then I will do the release, and update the platform

kdubb:
👍 All good for me

kdubb:
@vsevel I think we can remove the compiler source/target versions altogether because they are set in the base Quarkivers POM; I noticed they are absent in my other Quarkiverse projects.

@gastaldi Can you confirm this?

Issue #249, opened 2024-03-20
Creating a transit secret engine key in 4.0.0-alpha.2 throws a VaultClientException with status 204


vsevel:
204 should be a success in that situation.

seems to be related to:

```
  public VaultRequest<VaultSecretsTransitCreateKeyResult> createKey(String mountPath, String name,
      VaultSecretsTransitCreateKeyParams params) {
    return VaultRequest.post(getTraceOpName("Create Key"))
        .path(mountPath, "keys", name)
        .body(params)
        .expectOkStatus()
        .build(VaultLeasedResultExtractor.of(VaultSecretsTransitCreateKeyResult.class));
  }
```

the old code was:
```
    public Uni<Void> createTransitKey(VaultClient vaultClient, String token, String keyName, VaultTransitCreateKeyBody body) {
        return vaultClient.post(opName("Create Key"), "transit/keys/" + keyName, token, body, 204);
    }
```

kdubb:
The old code had an escape hatch for when Vault decided to produce warnings for a normally `204` method. When it did this it would return a `200` instead of `204` and include only the warnings in the response. The old code would basically allow any `204` to be promoted to a `200` and ignore anything but warnings. I think that is probably how the old code worked.

Transit's "Create Key" certainly does seem to normally return a `200` and the info for the created key.

kdubb:
That was a bit misleading... the old behavior doesn't exist but warning extraction happens well before interpreting the status code. So the promotion from 204 -> 200 is no longer necessary.

I believe the old "Create Key" was relying on this promotion from 204 - 200 and ignoring the result. From the client test suite, it seems clear that "Create Key" returns a 200 and information about the newly created key.

vsevel:
did not happen with the previous vault extension version. I think we need to do something about it. 

``` 
VaultClientException{operationName='VAULT [SECRETS (transit)] Create Key', requestPath='https://.../v1/transit/keys/...', status=204}
	at io.quarkus.vault.client.http.VaultHttpClient.throwVaultException(VaultHttpClient.java:80)
	at io.quarkus.vault.client.http.VaultHttpClient.lambda$buildResponse$0(VaultHttpClient.java:33)
	at java.base@21.0.2/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:684)
	at java.base@21.0.2/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662)
[...]

PR #251, opened 2024-03-20
Make Transit key creation return an Optional key to support previous Vault versions

kdubb:


kdubb:
@vsevel This should fix #249

kdubb:
I made it return an Optional as discussed.

client/src/main/specs/secrets/transit.yaml
```diff
@@ -32,7 +33,8 @@ operations:
       type: String
   result:
     kind: leased
-    unwrapData: true
+    unwrapUsing: "r == null ? Optional.empty() : Optional.of(r.getData())"
```
vsevel, line 36: `Optional.ofNullable(r)`

vsevel (approved):

client/src/main/specs/secrets/transit.yaml
kdubb: Good catch

PR #254, opened 2024-03-26
Allow applying a read timeout to JDKVaultHttpClient

### LINKED ISSUE #253: Read timeout is not set on Vault Requests
In `VaultSecretsKV1RequestFactory`, requests are created such as:

```
  public VaultRequest<VaultSecretsKV1ReadResult> read(String mountPath, String path) {
    return VaultRequest.get(getTraceOpName("Read"))
        .path(mountPath, path)
        .expectOkStatus()
        .build(VaultLeasedResultExtractor.of(VaultSecretsKV1ReadResult.class));
  }
```

we do not set read timeout on those requests.
the value should come from the `VaultRuntimeConfig.readTimeout`
as a result the value is always `Duration timeout = Duration.ofSeconds(30);` from `VaultRequest.Builder`.
as a side not, the defaults changed as well. In `VaultRuntimeConfig` the default is `5 secs`, whereas in `VaultRequest`, the default is `30 secs`. Users will incur a change of behavior even if the did not set a read timeout.

kdubb:
Read timeout was previously set on the VertxVaultHttpClient client but not on JDK. This applies the read timeout to JDKVaultHttpClient

Fixes #253

### DIFF
--- a/client/src/main/java/io/quarkus/vault/client/common/VaultRequest.java
+++ b/client/src/main/java/io/quarkus/vault/client/common/VaultRequest.java
@@ -368,6 +368,8 @@ public class VaultRequest<T> {
         var builder = new Builder<T>(operation, method);
         builder.baseUrl = baseUrl;
         builder.apiVersion = apiVersion;
+        builder.operation = operation;
+        builder.method = method;
         builder.path = path;
         builder.token = token;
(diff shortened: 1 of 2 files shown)


vsevel:
I have tested that the timeout the is applied indeed to the jdk http request.

PR #263, opened 2024-04-09
Update `VaultToken.cached` to only update the `cached` field and nothing else

### LINKED ISSUE #262: VaultToken.cached incorrectly extends the tokens lifetime
This should only update the `cached` field but inadvertently updates the parent `VaultTimeLimited.created` field as well.

This creates tokens with expiration times that do not agree with Vault; therefore, the client can attempt to use expired tokens.

From discussion #261

kdubb:
Fixes #262

### DIFF
--- a/client/src/main/java/io/quarkus/vault/client/auth/VaultTimeLimited.java
+++ b/client/src/main/java/io/quarkus/vault/client/auth/VaultTimeLimited.java
@@ -15,8 +15,12 @@ public abstract class VaultTimeLimited {
     public Duration leaseDuration;
 
     public VaultTimeLimited(boolean renewable, Duration leaseDuration, InstantSource instantSource) {
+        this(renewable, leaseDuration, instantSource.instant(), instantSource);
+    }
+
+    public VaultTimeLimited(boolean renewable, Duration leaseDuration, Instant created, InstantSource instantSource) {
         this.instantSource = instantSource;
-        this.created = instantSource.instant();
+        this.created = created;
         this.renewable = renewable;
         this.leaseDuration = leaseDuration;
     }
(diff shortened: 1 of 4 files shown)


client/src/main/java/io/quarkus/vault/client/auth/VaultTimeLimited.java
wiebeck, line 1: Consider marking all fields of `VaultTimeLimited` `final`. These objects are supposed to be immutable, right?

client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
```diff
@@ -31,8 +32,15 @@ public VaultToken(String clientToken, boolean renewable, Duration leaseDuration,
...
+    public VaultToken(String clientToken, boolean renewable, Duration leaseDuration, Instant created, boolean fromCache,
+            InstantSource instantSource) {
+        super(renewable, leaseDuration, created, instantSource);
+        this.clientToken = clientToken;
+        this.fromCache = fromCache;
+    }
+
     public VaultToken cached() {
-        return new VaultToken(clientToken, renewable, leaseDuration, true, instantSource);
+        return new VaultToken(clientToken, renewable, leaseDuration, created, true, instantSource);
```
wiebeck, line 43: Please add a test case for (at least) this method:
```java
@Test
void cachedKeepsCreatedTimestamp() {
    VaultToken t1 =
        VaultToken.from("foo", true, Duration.ofMinutes(1), tickableInstanceSource);
    tickableInstanceSource.tick(Duration.ofSeconds(10));
    VaultToken t2 = t1.cached();
    assertEquals(t1.getExpiresAt(), t2.getExpiresAt());
}
```
(this would fail in v4.0.0)

client/src/main/java/io/quarkus/vault/client/auth/VaultTimeLimited.java
kdubb: I wrapped the fields and made them final.

client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
kdubb: I added a higher-level test that ensures cached tokens are the same as those returned from the original token provider (except the cached flag).

vsevel:
LGTM

PR #257, opened 2024-03-28
Map HTTP timeout exceptions to TimeoutException

kdubb:

PR #280, opened 2024-05-22
First time fetch in Vault Config Source done multiple times

### LINKED ISSUE #278: First time fetch in Vault Config Source done multiple times
This is related to this:
```
        if (firstTime) {
            log.debug("fetch secrets first time with attempts = " + vaultRuntimeConfig.mpConfigInitialAttempts());
            fetchSecretsFirstTime(properties);
            firstTime = false;
```
If there is a failure, `firstTime` is still true, and next attempt will be done as if this was the first time.
As a result, the application repeats multiple times, several load attempts, delaying the eventual failure of application startup.

vsevel:
Fixes https://github.com/quarkiverse/quarkus-vault/issues/278

moved `firstTime = false;` above the call to `fetchSecretsFirstTime`
added a finally block on `cache.set(new VaultCacheEntry<>(properties));`
caught `HttpTimeoutException`

to reproduce, used toxiproxy with:
```
toxiproxy-cli create -l localhost:18200 -u localhost:8200  vault
toxiproxy-cli toxic add -t latency -a latency=5000 -a jitter=2000 vault
```

and configuration:
```
quarkus.vault.url=http://localhost:18200
quarkus.vault.mp-config-initial-attempts=3
quarkus.vault.secret-config-kv-path=myapps/vault-quickstart/config,myapps/vault-quickstart/config2
```

### DIFF
--- a/client/src/main/java/io/quarkus/vault/client/http/jdk/JDKVaultHttpClient.java
+++ b/client/src/main/java/io/quarkus/vault/client/http/jdk/JDKVaultHttpClient.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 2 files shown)


kdubb:
@vsevel With #257 you should be catching `TimeoutException`, correct?

vsevel:
Well no. I tested. That IS what we are getting. Or we need to convert it when it is raised somewhere

kdubb:
Then #257 wasn't implemented correctly. I think we should be fixing that problem rather than attempting to catch multiple timeouts all over the place; it's just easier to remember.

vsevel:
Ok I will give it a try. It just I feel I have been chasing that one up so many times...

kdubb:
I'd leave it as is or maybe make it recursive to make it easier to add exceptions to. Of course, then you're relying on `Throwable.cause` not being recursive, and I think it sometimes is.

vsevel:
> make it recursive to make it easier to add exceptions to. Of course, then you're relying on Throwable.cause not being recursive, and I think it sometimes is.

that is what I was concerned of. I think we can live with the repetition. I will squash then if you are ok.

kdubb:
👍

PR #286, opened 2024-06-12
Remove the usage of TLSConfig

cescoffier:
TLSConfig became a runtime config and cannot be injected anymore.


cescoffier:
Closing in favor of #285

PR #316, opened 2024-08-22
Allow configuring different images for Postgres and RabbitMQ in testing

### LINKED ISSUE #315: Allow configuring different images for Postgres and RabbitMQ in testing
We have an issue where on the test RHEL 9.2 machine, `rabbitmq:3.7.25-management-alpine` won't run. It doesn't print anything. However, `rabbitmq:3.8.4-management-alpine` works. But there is no way how to configure the image for testcontainers integration in the https://github.com/quarkiverse/quarkus-vault/blob/main/test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java.

I would like to implement optional override of the default images for testing.

Would you also be able to release 3.5.x release with this (I will also provide a PR if there is a branch) for us?

xstefank:
Fixes #315 

Let me know if there is a place where to test this.

### DIFF
--- a/test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
+++ b/test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
@@ -33,6 +33,9 @@ import java.util.regex.Pattern;
 
 import javax.sql.DataSource;
 
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logging.Logger;
 import org.testcontainers.containers.Container;
 import org.testcontainers.containers.GenericContainer;
@@ -40,6 +43,7 @@ import org.testcontainers.containers.Network;
 import org.testcontainers.containers.PostgreSQLContainer;
 import org.testcontainers.containers.RabbitMQContainer;
 import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.utility.DockerImageName;
 
 import io.quarkus.vault.VaultKVSecretEngine;
 import io.quarkus.vault.client.VaultClient;
@@ -98,6 +102,13 @@ public class VaultTestExtension {
     public static final String OUT_FILE = "/out";
     public static final String WRAPPING_TEST_PATH = "wrapping-test";
 
+    public static final String TEST_POSTGRES_CONFIG = "vault-test.postgres.version";
+    public static final String TEST_RABBITMQ_CONFIG = "vault-test.rabbitmq.version";
+    public static final String TEST_RABBITMQ_4_CONFIG = "vault-test.test-rabbitmq-4";
+    public static final String TEST_POSTGRES_VERSION = PostgreSQLContainer.DEFAULT_TAG;
+    public static final String TEST_RABBITMQ_VERSION_3 = "3.13.7-management-alpine";
+    public static final String TEST_RABBITMQ_VERSION_4 = "4.0.2-management-alpine";
+
     private static final String CRUD_PATH = "crud";
 
     public GenericContainer vaultContainer;
@@ -114,6 +125,8 @@ public class VaultTestExtension {
 
     private String db_default_ttl = "1m";
     private String db_max_ttl = "10m";
+
+    private Config config = ConfigProvider.getConfig();
 
     public static void testDataSource(DataSource ds) throws SQLException {
         try (Connection c = ds.getConnection()) {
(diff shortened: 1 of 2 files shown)


test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
```diff
@@ -201,7 +207,9 @@ public void start() throws Exception {
 
         postgresContainer.setPortBindings(Arrays.asList(MAPPED_POSTGRESQL_PORT + ":" + POSTGRESQL_PORT));
 
-        rabbitMQContainer = new RabbitMQContainer()
+        // 3.7.25-management-alpine is the default version in RabbitMQContainer but the constants is private
```
vsevel, line 210: should we access it through reflection? if not we are going to have to make sure we stay aligned (and it is probably going to be forgotten).

vsevel:
> Let me know if there is a place where to test this.

what do you mean? if somebody has a project where he uses rabbit or postgres?

kdubb:
I'm not sure what problem this is fixing exactly. It seems you are developing this extension on a specific RHEL version that doesn't run our test version?

kdubb:
To be more clear... the `VaultTestExtension` is intended only for running tests for this extensions development. Are you using it for extension development or using it in another project to test that?

xstefank:
Using it to run this part of quarkus-platform testing - https://github.com/quarkusio/quarkus-platform/tree/main/generated-platform-project/quarkus/integration-tests/quarkus-vault-integration-test. It fails on RHEL 9.2 because rabbitmq won't start. I tested it on this machine and it won't print anything so I have no idea what's going on. This fix will allow us to bump the rabbitmq version to something that will work on that machine.

@vsevel Updated it to use reflection. I asked whether you would like to have this tested but I don't think it's necessary.

vsevel:
just like @kdubb was saying, we had no idea this was used outside this project. as such, considering we would accept this PR, when we make changes to it for the needs of this extension, we have no idea whether this would break something in the platform. are we ok with that? cc @gsmet

xstefank:
unfortunately, this is already latest - see https://github.com/testcontainers/testcontainers-java/blob/1.20.1/modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQTestImages.java and https://central.sonatype.com/artifact/org.testcontainers/rabbitmq/1.20.1/versions

xstefank:
hey @vsevel and @kdubb how do you think we should proceed here?

vsevel:
I think we can proceed. what do you think @kdubb ?

vsevel:
I am away from my desk. We will finalize when I come bac in approx 2 weeks

test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
kdubb: That's a really old, unsupported version of RabbitMQ. Community support for RMQ 3.13 just ended. Maybe we should just be pestering TestContainers to get off their asses 🤣

kdubb:
On one hand, I want to help; on the other, this seems like we're opening up something that we expect to remain closed.

I don't want to be rigid, but I believe both @vsevel and I already have lots on our plates, and I want to make sure we're not adding to maintenance. However, I don't really see the harm here.

Another option would be to pull and retag 3.X to 3.7.25 in your test machine's docker, assuming the test machine/environment is ephemeral.

kdubb:
After pressing "Comment" I'm realizing, we should absolutely not be testing Vault with RabbitMQ 3.7. So if TestContainers is going to let severely out-of-date (and out of support) version be the default. We definitely need to define a version and update it ourselves.

If what I just said is the actual situation, and I'm not missing something, then @xstefank would you mind changing the PR to define constants for RMQ and PG?

Given that RMQ is now split on 4.x and 3.x, we should probably be testing both of those release lines (given that major changes for them are few and far between).

xstefank:
@kdubb keeping the override, or will you be taking care of updating the versions?

test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
```diff
@@ -201,7 +214,12 @@ public void start() throws Exception {
 
         postgresContainer.setPortBindings(Arrays.asList(MAPPED_POSTGRESQL_PORT + ":" + POSTGRESQL_PORT));
 
-        rabbitMQContainer = new RabbitMQContainer()
+        rabbitMQContainer = new RabbitMQContainer(
+                ((DockerImageName) FieldUtils.readStaticField(RabbitMQContainer.class, "DEFAULT_IMAGE_NAME", true)).withTag(
+                        config.getOptionalValue(TEST_RABBITMQ_CONFIG, String.class)
+                                .orElse(config.getOptionalValue("test-rabbitmq-4", Boolean.class).orElse(Boolean.FALSE)
```
vsevel, line 220: why not use the `vault-test` prefix that you used on the other properties?

vsevel:
@kdubb do you still have some concerns about the approach?

test-framework/src/main/java/io/quarkus/vault/test/VaultTestExtension.java
xstefank: thanks @vsevel, updated

kdubb:
I'm fine with this.

vsevel:
Thanks @xstefank

Issue #327, opened 2024-09-27
NullPointerException when retrieving Vault contents with null values


joeyy-watts:
This issue is found in version `4.0.1`, although I still see the problematic code in `4.1.0`.

### Issue

In my use case, I have a JSON secret in Vault which can contain null values, e.g:

```
{
     "password": "foo",
     "someotherfield": null
}
```

When I try to read this secret with `quarkus-vault`, the following exception is thrown:

```
2024-09-26 12:45:19,744 WARN  [io.agr.pool] (agroal-21) Datasource '<default>': Failed to create connection due to NullPointerException
2024-09-26 12:45:19,744 ERROR [io.sma.health] (executor-thread-1) SRHCK01000: Error processing Health Checks: java.lang.NullPointerException
    at java.base/java.util.Objects.requireNonNull(Unknown Source)
    at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Unknown Source)
    at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(Unknown Source)
    at java.base/java.util.Iterator.forEachRemaining(Unknown Source)
    at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
    at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)
    at java.base/java.util.stream.ReferencePipeline.collect(Unknown Source)
    at io.quarkus.vault.runtime.VaultKvManager.convert(VaultKvManager.java:38)

```

On [this line in VaultKvManager.java](https://github.com/quarkiverse/quarkus-vault/blob/main/runtime/src/main/java/io/quarkus/vault/runtime/VaultKvManager.java#L38,) I've found the cause to be the way the library is casting the JSON value to String.

```
(String) entry.getValue()
```

In the case that the entry's value is `null` this will also return `null`, causing the exception above.

I suspect it will also fail if the value is of a non-String type, with the following exception (although I haven't tested with the library; just ran some scratch code):

```
[...]

PR #329, opened 2024-09-28
use .readSecretJson in VaultCredentialsProvider

joeyy-watts:
Fixes https://github.com/quarkiverse/quarkus-vault/issues/327

I have tested this in my own fork of the repo, and confirmed it to be working.

---

Under `.getCredentials()`, the current code uses `.readSecret()` which will fail if the secret containing the password contains a null value, e.g.:

```
{
    "password": "foo",
    "someotherfield": null
}
```

This PR uses `.readSecretJson()` instead, which does not assume the secret's values are always String, and converts it to String using `String.valueOf()`.

One caveat is that it will no longer (implicitly) guarantee the password to never be null. I am open to suggestions on this point.

### DIFF
--- a/runtime/src/main/java/io/quarkus/vault/runtime/VaultCredentialsProvider.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/VaultCredentialsProvider.java
@@ -56,7 +56,7 @@ public class VaultCredentialsProvider implements CredentialsProvider {
         }
 
         if (config.kvPath().isPresent()) {
-            String password = vaultKVSecretEngine.readSecret(config.kvPath().get()).get(config.kvKey());
+            String password = String.valueOf(vaultKVSecretEngine.readSecretJson(config.kvPath().get()).get(config.kvKey()));
             Map<String, String> result = new HashMap<>();
             result.put(PASSWORD_PROPERTY_NAME, password);
             return result;


kdubb:
@vsevel You're in a better position to determine if this is the best fix for this.

runtime/src/main/java/io/quarkus/vault/runtime/VaultCredentialsProvider.java
```diff
@@ -56,7 +56,7 @@ public Map<String, String> getCredentials(String credentialsProviderName) {
         }
 
         if (config.kvPath().isPresent()) {
-            String password = vaultKVSecretEngine.readSecret(config.kvPath().get()).get(config.kvKey());
+            String password = String.valueOf(vaultKVSecretEngine.readSecretJson(config.kvPath().get()).get(config.kvKey()));
```
vsevel, line 59: this is fine. I would just add a check for nullity.
```
var val = vaultKVSecretEngine.readSecretJson(config.kvPath().get()).get(config.kvKey());
if(val == null) {
  throw new VaultException("unable to retrieve credential "+config.kvKey()+" from path "+config.kvPath().get());
}
```

vsevel:
superseded by https://github.com/quarkiverse/quarkus-vault/pull/349

PR #345, opened 2024-12-02
Vault Transit Data Key

vsevel:
support data key generation, plus add a test for key rotation


vsevel:
any chance you could review this @kdubb ?

kdubb (approved):

PR #360, opened 2025-01-23
Support refresh of limited use tokens (num_use > 0)

### LINKED ISSUE #338: No support for token_num_uses in auth token
Hi,

we stumbled across an issue where `token_num_uses` is not supported in the vault extension.
When you have a token that has a `token_num_uses` value of anything else other than `0`, and it reaches the limit, it does not automatically renew the token.

So any further requests fail with `403 Not Authorized` and a restart of the application is necessary or you need to wait until the TTL expires, and a new token is created.

Thank you!

kdubb:
Fixes #338

While enhancing the test for caching tokens I realized supporting limited use tokens should be fairly straightforward.

@vsevel What do you think? The only tricky thing is that we copy token values (e.g., for caching) so I used an `AtomicReference` as a shared reference count. Other than that, I added `getClientTokenForUsage` to decrement the count when it's being used for authorization and overloaded the expiration tests to take into account it's max # of uses.

I added a test to the caching token provider tests suite that tests limited use tokens and it seems to perform as expected.

It does not attempt to `extend` limited use tokens that are over their use limit because from my investigation it doesn't seem that's allowed.

@alexjaravete Can you build this this locally and see if this "just works" for you?

### DIFF
--- a/client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
+++ b/client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
@@ -1,10 +1,12 @@
 package io.quarkus.vault.client.auth;
 
 import static io.quarkus.vault.client.logging.LogConfidentialityLevel.LOW;
+import static java.lang.Math.max;
 
 import java.time.Duration;
 import java.time.Instant;
 import java.time.InstantSource;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import io.quarkus.vault.client.logging.LogConfidentialityLevel;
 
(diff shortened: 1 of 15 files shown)


vsevel:
ideally we would want decrementing the token close to where it is actually being used, so in the real `VaultRequestExecutor`, to make sure it is actually called exactly once. `VaultClient` is an executor, but so are `JDKVaultHttpClient`, `VertxVaultHttpClient`. are we sure we always go through `VaultClient` first, then to one of the concrete classes? it seems so. 
for my own comprehension what are you supposed to do when you exhausted the number of uses? you can still renew it?
so when we call `getClientTokenForUsage` and it fails with a `VaultException` do we auto renew it? 
I see the overload on `isExpired()`, so the sequence what should happen is that when we reaches 0, we should renew the token. correct?
I suppose the only pitfall would be if there was 1 use left, with 2 concurrent calls. they would both decrement the counter, after they have called `isExpired()`, then fail when making the real call. is that correct?
if true, then this approach would not be reliable in highly concurrent applications.
a solution then would be that once `isExpired()` has returned `false`, we can't be rejected because of the num uses.
but that might not be easily implementable?

kdubb:
> VaultClient is an executor, but so are JDKVaultHttpClient, VertxVaultHttpClient. are we sure we always go through VaultClient first, then to one of the concrete classes? it seems so.
 
Yes all roads go through `VaultClient`, it basically wraps another executor that is the client implementation. The only calls to `getClientToken` were the one in `VaultClient` and one in `VaultCachingTokenProvider`. I added `getClientTokenForUsage` only to be very clear about what it's doing.

> I suppose the only pitfall would be if there was 1 use left, with 2 concurrent calls. they would both decrement the counter, after they have called isExpired(), then fail when making the real call. is that correct?

This is a great point. I'll look into solving this problem and testing it.

alexjaravete:
@kdubb Thanks for taking a look at it and creating a fix. I will try to run it locally in the next few days and let you know if it works.

alexjaravete:
I did a small test in our system with the new fix, and it seems to work as expected. The potential pitfall mentioned by @vsevel could still be a problem, but I was not able to test this in my limited test run.
And thank you @kdubb for tackling this issue.

vsevel:
@kdubb I would like to do a release for the next quarkus version.
do you think this could go in?

vsevel:
@kdubb have you been able to make any progress? worst case scenario, can the current non perfect fix be considered as an improvement, and we improve on the concurrency later?

client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
```diff
@@ -5,42 +5,52 @@
...
 
-    public VaultToken(String clientToken, boolean renewable, Duration leaseDuration, boolean fromCache,
+    public VaultToken(String clientToken, boolean renewable, Duration leaseDuration, Integer allowedUsesRemaining,
+            boolean fromCache,
             InstantSource instantSource) {
         super(renewable, leaseDuration, instantSource);
         this.clientToken = clientToken;
         this.fromCache = fromCache;
+        this.allowedUsesRemaining = new AtomicInteger(
+                allowedUsesRemaining != null && allowedUsesRemaining > 0 ? allowedUsesRemaining : Integer.MAX_VALUE);
```
vsevel, line 40: why this condition `allowedUsesRemaining > 0` ?
if `allowedUsesRemaining <= 0`, then the remaining allowed uses should really be `MAX_VALUE`?

vsevel:
I am planning a vault release end of this week with `3.19`. it would be nice to finalize this PR

kdubb:
@vsevel Apologies for the quietness been slammed with other things. I do think this an improvement either way because limited use tokens are more likely to used in a managed way. By that I mean that you *probably* are not using a limited use token as the "general" authorization token. Maybe @alexjaravete can shed some light here.

In either case I'm looking into this a bit today to see what can be done to mitigate this issue.

kdubb:
@vsevel @alexjaravete I did a deep dive and added some extensive tests and I think the current implementation mitigates the race condition well.

### Token Acquisition/Use
The two step process token acquisition and then application naturally creates a race condition regardless of caching. I don't really see how we could change this it's just worth pointing out that the client allows _a lot_ of flexibility in how a token is acquired (even possibly creating it on the fly). Given this, we'll always have a race condition to deal with.

### Execution is a "use"
This implementation treats any attempted execution of a request as a "use" of the token. Generally, that means a _sent_ HTTP request but considering the allowed flexibility in `VaultClient` (it just passed the request to an executor) there could potentially be something else fail before ever sending a request. Even in the common case of passing the request directly to an HTTP executor, a network failure would still incorrectly consume a token usage.

I don't see this as a problem. It would only mean that a new token is being requested while the current token still has a couple uses left. Maybe somebody else has a different scenario that would mean this is not an allowable state.

### Over "use" is properly handled
Under normal operation tokens are cached and prior to usage their expiry and usage count are  checked before being applied. If either case fails an extension or refresh is attempted. In the race condition case, a `VaultTokenException` (newly added) is thrown and `VaultClient` handles this by retrying the request, with a call to the token provider first. Currently, the default retry count is `2` making the max number of attempts `3` for any request; although this is now configurable in `VaultClient`.

In some bizarre case of multiple clients, that are not sharing a token provider, using a limited token, the server returns the error and the client initiates a retry as expected.

### Other Considerations

Token over "use" could potentially be enhanced. For example, we could "lease" a token `use` to requests and only permanently decrement the token's remaining use count once a request has been returned from the Vault instance. Part of request execution would have to be waiting for the lease of a `use`. This would essentially mean some kind of serialization *might* happen as tokens near usage exhaustion; which sounds correct if we believe sending a possibly overused token should always be avoided.

Alternatively, we could yank the checks for over use altogether and let the server tell us when to refresh (via a 403 response, which should now be appropriately retried).

### Wrap Up

All the above suggests to me that the current implementation is "good". If a token over use is detected by the client, or server, it is retried up to 2 times. A high volume client can increase the `maxRetries` in `VaultClient` if they expect a lot of requests when using a limited use token.

We could up the default # of retries as well because they only apply to token & authentication failures. Separately (in another PR) we should probably add a general "retry" count for things like network failures.

So.. unless I'm missing something... 👍👍

kdubb:
@vsevel I switched around the `getClientTokenForUsage` to check the result of the `updateAndGet` instead of using a pre-check using `hasAllowedUsesRemaining`.

It's more correct and I believe that's what you were suggesting.

Should we merge now?

client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
kdubb: The value is either `0` (which is unlimited) or the number of allowed uses, I believe the null check was just me being pedantic.

In either case we have to map `0`/`null` to something or we'd have to check `allowedUsesRemaining != null` before every access. This seemed like the cleaner method.

kdubb:
@vsevel I responded to the review inline.

client/src/main/java/io/quarkus/vault/client/auth/VaultToken.java
vsevel: I can see it now in https://developer.hashicorp.com/vault/api-docs/auth/token#num_uses
seemed strange that 0 uses means unlimited.

vsevel:
@alexjaravete would you care to do one more test to make sure it works for you?

vsevel (approved):

alexjaravete:
@vsevel Sorry I was not able to test it anymore, but I will test it with the new release. Thank You!

vsevel:
I am going to try to release before the end of the week

Issue #328, opened 2024-09-27
Question: Why is default value of token refresh 30 seconds?


Visserr2:
Question: Why is default value of token refresh 30 seconds?

![{065DE0E0-FF39-4EAC-B7BF-AC51C0D2B2C3}](https://github.com/user-attachments/assets/f4e44420-750b-4e49-aec6-810f1e55a36f)

Is there a way to override this (I cannot find it) ? In practice always a new token will be created (unless there is a call triggered within the 30 seconds of the expiry time.

kdubb:
That's configurable in the `VaultClient` I'm not sure if it's configurable via Quarkus config. I'd need to look it up.

More importantly, that's not a token refresh period. IF a token expires within the next "renew grace period" then the token will be refreshed. So if your token TTL is 5 minutes, and you have the default of 30 seconds grace period, your token will be updated after 4 minutes and 30 seconds.

It's not a configuration you should ever really need, unless your Vault instance is under such load that it takes 20+ seconds to respond to token renewals.

Visserr2:
Hmm, that is not how I experience it.  I tried it with following config (as test):


Token Expiry : 2minutes
max-lifetime: 30s
Grace period: 1m
-----------------------------------------------------------------
2024-09-28 00:11:44,654 FINE  [io.qua.vau.cli.aut.VaultCachingTokenProvider] (agroal-11) using cached token {clientToken: ***, renewable: true, leaseDuration: PT2M, valid_until: 2024-09-27T22:13:10.923708300Z} (expires at 2024-09-27T22:13:10.923708300Z)
2024-09-28 00:12:15,149 FINE  [io.qua.vau.cli.aut.VaultCachingTokenProvider] (agroal-11) using cached token {clientToken: ***, renewable: true, leaseDuration: PT2M, valid_until: 2024-09-27T22:13:10.923708300Z} (expires at 2024-09-27T22:13:10.923708300Z)
2024-09-28 00:12:45,604 FINE  [io.qua.vau.cli.aut.VaultCachingTokenProvider] (agroal-11) using cached token {clientToken: ***, renewable: true, leaseDuration: PT2M, valid_until: 2024-09-27T22:13:10.923708300Z} (expires at 2024-09-27T22:13:10.923708300Z)
[...]

PR #361, opened 2025-01-23
Explore grace period renewal (#328)

kdubb:
This PR is currently just to get a test reproducing the issue described in #328.


kdubb:
I've updated `testExpiringTokensRenew` to simulate a 1 hour token and test that the grace period of 30 seconds causes the token to renew after `59:30`.

vsevel:
Closing in favor of #526, which reproduces the issue described in #328 both at the token-renewal level (approle vs userpass, against a Vault stub) and end-to-end with dynamic database credentials against a real Vault + PostgreSQL: on current main both approle scenarios fail while the identical userpass scenarios pass. The one-line fix is proposed in #527.

PR #367, opened 2025-02-10
Actually pass standby-ok config params to health check

wiebeck:
In #320 I already fixed an issue with letter casing for the two health check query params `standbyok` and `perfstandbyok`, now we noticed that these two options are never ever being used in production. `VaultHealthCheck#call` only calls the parameter-less `status()` method instead of the method accepting two parameters for `standbyok` and `perfstandbyok`, so the config parameters were actually ineffective altogether.

This PR fixes this behavior.

### DIFF
--- a/runtime/src/main/java/io/quarkus/vault/runtime/health/VaultHealthCheck.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/health/VaultHealthCheck.java
@@ -27,13 +27,10 @@ public class VaultHealthCheck implements HealthCheck {
         final HealthCheckResponseBuilder builder = HealthCheckResponse.named("Vault connection health check");
 
         try {
-            Boolean standbyok = null;
-            Boolean perfstandbyok = null;
-            if (buildTimeConfig.health() != null) {
-                standbyok = buildTimeConfig.health().standByOk();
-                perfstandbyok = buildTimeConfig.health().performanceStandByOk();
-            }
-            var status = client.sys().health().status(standbyok, perfstandbyok)
+            boolean isStandByOk = this.buildTimeConfig.health().standByOk();
+            boolean isPerfStandByOk = this.buildTimeConfig.health().performanceStandByOk();
+
+            var status = client.sys().health().status(isStandByOk, isPerfStandByOk)
                     .toCompletableFuture().get();
 
             switch (status) {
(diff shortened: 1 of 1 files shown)


wiebeck:
I would love to provide some tests but I'm having problems setting up a proper `@QuarkusTest` with mocked configuration in order to verify the correct method being called.

vsevel:
I will look at this today

runtime/src/main/java/io/quarkus/vault/runtime/health/VaultHealthCheck.java
```diff
@@ -17,13 +18,22 @@ public class VaultHealthCheck implements HealthCheck {
...
+
     @Override
     public HealthCheckResponse call() {
 
         final HealthCheckResponseBuilder builder = HealthCheckResponse.named("Vault connection health check");
 
         try {
-            var status = client.sys().health().status()
+            Boolean standbyok = null;
+            Boolean perfstandbyok = null;
```
wiebeck, line 31: I've chosen nullable objects over primitives because the health client won't even add query parameters if they are passed as `null`.

vsevel: I do not think there is need for this.
those fields had been added for `VaultSystemBackendReactiveEngine.health()`

and in the impl we can see
```
    @Override
    public Uni<VaultHealth> health() {
        boolean isStandByOk = this.buildTimeConfig.health().standByOk();
        boolean isPerfStandByOk = this.buildTimeConfig.health().performanceStandByOk();

        return this.health(isStandByOk, isPerfStandByOk);
    }
```
there is no protection against nullity, and we have defaults for those 2 attributes.
can you check? other than that this looks good.

wiebeck: Adapting my code to what already exists in `VaultSystemBackendManager`.

vsevel:
did you re-test?

vsevel:
tested on my side. thanks @wiebeck

vsevel:
@wiebeck more specifically I tested that without the config params, this is still working as before (no regression).
I will you let check that with the params in the config you get the new behavior.

PR #381, opened 2025-03-14
Support for multiple mount paths selectable with prefixes in kv engine and ConfigSource

plevart:
Hi,

It is usual practice to organize secrets in vault using two or more mount paths. For example, one monut path for secrets shared among multiple applications, another mount path for secrets private to the application, etc...
Currenty it is imposible to configure quarkus-vault to retrive secrets using multiple mount paths or map them to config properties from multiple mount paths in the same app.
I therefore propose to extend the quarkus-vault with this possibility in a backward compatible way.
My proposal consists of the addition of the following map of configuration properties:

`quarkus.vault.kv-secret-engine-mount-path."prefix"`

Similar to existing map of `quarkus.vault.secret-config-kv-path."prefix"` properties, but it contains a single mount path per `prefix`. These mount paths, if configured, are used instead of default mount path configured with  `quarkus.vault.kv-secret-engine-mount-path` property when using the overloaded VaultKVSecret(Reactive)Engine methods that take the `prefix` as the 1st additional parameter, or when mapping secrets to config propertes in ConfigSource for paths configured with map of configuration properties:

`quarkus.vault.secret-config-kv-path."prefix"`
 
The following is the rule that pairs a path from `quarkus.vault.secret-config-kv-path."prefix"` with the mount path from `quarkus.vault.kv-secret-engine-mount-path."prefix"`:
 
Let's call the prefix in `quarkus.vault.secret-config-kv-path."prefix"` the `configPrefix` and let's call the prefix in `quarkus.vault.kv-secret-engine-mount-path."prefix"` the `mountPrefix`. The mount path used in combination with the config path is the one with the longest `mountPrefix` that is a prefix of the `configPrefix`, if any such exists or the default mount path configured with `quarkus.vault.kv-secret-engine-mount-path`.

For example, with the following configuration:

```
quarkus.vault.kv-secret-engine-mount-path."shared"=secret/shared
quarkus.vault.kv-secret-engine-mount-path."app1"=secret/app1

quarkus.vault.secret-config-kv-path."shared.kafka"=kafka
quarkus.vault.secret-config-kv-path."app1.db"=db
```

The keys from path (assuming v2 engine):

`/v1/secret/shared/data/kafka`

...will end up as config properties prefixed with `shared.kafka.*` and keys from path:

`/v1/secret/app1/data/db`

...will end up as config properties prefixed with `app1.db.*`

I came up with code changes to implement this feature and are contained in this PR. The start of them, of course, just to trigger some feedback.

So, WDYT?

Regards, Peter

### DIFF
--- a/integration-tests/vault/src/test/java/io/quarkus/vault/VaultMultiMountPathConfigITCase.java
+++ b/integration-tests/vault/src/test/java/io/quarkus/vault/VaultMultiMountPathConfigITCase.java
@@ -1,0 +1,49 @@
+package io.quarkus.vault;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.vault.test.VaultTestLifecycleManager;
+
+@DisabledOnOs(OS.WINDOWS) // https://github.com/quarkiverse/quarkus-vault/pull/381
+@QuarkusTestResource(VaultTestLifecycleManager.class)
+public class VaultMultiMountPathConfigITCase {
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .withApplicationRoot((jar) -> jar
+                    .addAsResource("application-vault-multi-mount-path.properties", "application.properties"));
+
+    @Test
+    public void defaultMountPathAndDefaultPath() {
+        Config config = ConfigProviderResolver.instance().getConfig();
+        assertEquals("red", config.getValue("color", String.class));
+        assertEquals("XL", config.getValue("size", String.class));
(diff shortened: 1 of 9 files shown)


vsevel:
I will start looking at it this week

vsevel:
this would work:
```
quarkus.vault.secret-config-kv-path."mykvv1/v1"=bar
quarkus.vault.kv-secret-engine-mount-path.mykvv1=secret-v1
```
in that case, in `VaultConfigSource` :
```
    private Map<String, String> fetchSecrets(String path, String prefix) {

        Map<String, Object> secretJson = getVaultKVSecretEngine().readSecretJson(path).await().indefinitely();
```
we have `prefix=mykvv1/v1`. we could do some parsing there and extract `mykvv1` as an alias.
or `quarkus.vault.secret-config-kv-path."mykvv1->v1"=bar` would work too.
but `quarkus.vault.secret-config-kv-path."mykvv1:v1"=bar` does not work.

but then we would need to do something similar for the kv version: `quarkus.vault.kv-secret-engine-version.mykvv1=1` 
I am starting to think we may want to create an config for a named secret engine that would have its version, mount path, and secret config kv path, instead of trying to support named configs for each parameter.

vsevel:
something like this:
```
quarkus.vault.kv-secret-engine."myengine".version=1
quarkus.vault.kv-secret-engine."myengine".mount-path=secret-v1
quarkus.vault.kv-secret-engine."myengine".secret-config-kv-path=bar
quarkus.vault.kv-secret-engine."myengine".secret-config-kv-path."foo"=myfoo
```
this would be much clearer.

vsevel:
Fixed by https://github.com/quarkiverse/quarkus-vault/pull/411

PR #389, opened 2025-04-17
Add config property to configure transit secret engine mount path

akraxx:
This adds `quarkus.vault.transit-secret-engine-mount-path` config property.

Behaviour is the same as `quarkus.vault.kv-secret-engine-mount-path`.

### DIFF
--- a/docs/modules/ROOT/pages/vault-transit.adoc
+++ b/docs/modules/ROOT/pages/vault-transit.adoc
@@ -159,6 +159,12 @@ quarkus.vault.authentication.userpass.username=bob
 quarkus.vault.authentication.userpass.password=sinclair
 ----
 
+[NOTE]
+====
+By default Quarkus assumes that a _transit secret engine_ is mounted at path `transit/`.
+If that is not the case, please use properties `quarkus.vault.transit-secret-engine-mount-path` accordingly.
+====
+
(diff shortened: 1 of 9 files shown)


vsevel:
LGTM. we need a test.

vsevel:
I added some tests.
can you have a look @kdubb ? I had to change some of tests from `client.secrets().transit()`  to 
```
    @ConfigProperty(name = "quarkus.vault.transit-secret-engine-mount-path", defaultValue = "transit")
    String transitMountPath;
...
    client.secrets().transit(transitMountPath)
```

in other words, `client.secrets().transit()` should point to `transit` (the default value), and not the configured overridden mount path. correct?

@akraxx can you rebase on `main`  fix conflicts, and squash?

kdubb:
@vsevel `VaultClient` does not obey any of the Quarkus configuration; it uses the Vault defaults. So, yes, you would need to configure the custom path as you have done.

vsevel:
I think we are ready for this. can you squash and rebase? thanks @akraxx

akraxx:
Hello @vsevel , it's ok now, sry for the delay

vsevel (approved):

vsevel:
thanks @akraxx

PR #411, opened 2025-07-14
Support Multiple KV Secret Engines

vsevel:
this bring support named kv secret engines with their mount path, version and `secret-config-kv-path`.
for instance for a `v1` engine:
```
quarkus.vault.kv-secret-engine."myengine1".version=1
quarkus.vault.kv-secret-engine."myengine1".mount-path=secret-v1

quarkus.vault.kv-secret-engine."myengine1".secret-config-kv-path=foo
# and/or
quarkus.vault.kv-secret-engine."myengine1".secret-config-kv-path."myfoo"=foo
```
and for `v2`:
```
quarkus.vault.kv-secret-engine."myengine2".version=2
quarkus.vault.kv-secret-engine."myengine2".mount-path=secret

quarkus.vault.kv-secret-engine."myengine2".secret-config-kv-path=foo
# and/or
quarkus.vault.kv-secret-engine."myengine2".secret-config-kv-path."myfoo"=foo
```

the default engine (with mount path `secret` and version 2) is still available, and can be used alongside the named engines.
for instance this configuration will bring in the configuration context both path `secret/bar` and `secret-v1/foo`:
```
quarkus.vault.kv-secret-engine.secret-config-kv-path=bar

quarkus.vault.kv-secret-engine."myengine1".mount-path=secret-v1
quarkus.vault.kv-secret-engine."myengine1".secret-config-kv-path=foo
```

this is an alternate proposition to https://github.com/quarkiverse/quarkus-vault/pull/381

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -345,6 +345,11 @@ This value can be changed using property `quarkus.vault.config-ordinal`.
 Only scalar types are supported for secrets value from Vault.
 ====
 
+[NOTE]
+====
+It is possible to create named KV secret engines, as a way to fetch secrets from different mount paths.
+====
+
 == Programmatic access to the KV secret engine
 
 Sometimes secrets are retrieved from an arbitrary path that is known only at runtime through an application
--- a/integration-tests/vault/src/test/java/io/quarkus/vault/VaultKvSecretEngineV1AliasITCase.java
+++ b/integration-tests/vault/src/test/java/io/quarkus/vault/VaultKvSecretEngineV1AliasITCase.java
@@ -1,0 +1,36 @@
+package io.quarkus.vault;
+
+import static io.quarkus.vault.test.VaultTestExtension.SECRET_VALUE;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.test.common.QuarkusTestResource;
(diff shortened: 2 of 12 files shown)


plevart:
So this configures independent named "engines", containing associated mount path, version and config path(s). A couple of questions:

I'm mainly interested for this in the context of ConfigSource. So what property names are provided with such engine? Analogue to the unnamed engine:

config-path-name.key

Or are properties prefixed with engine name:

engine-name.config-path-name.key


What happens when there are conflicts when multiple engines define same property?

vsevel:
@kdubb can you take a look? ideally I would like this to go in `3.25` , which in `CR1` this wednesday (we have to release anyway because of https://github.com/quarkiverse/quarkus-vault/pull/407)
@plevart and @ousgougou can you make sure this works for you?

plevart:
When mutiple engines define same property, values will be overridden, but it is not defined which engine wins. Do you think a kind of ordering could be added to the engines? A config path property:

quarkus.vault.secret-config-kv-path=config/path1,config/path2

...may define multiple paths from which the key-value pairs are gathered and the order of path definitions governs which might override which on the property-key-level. The path defined "later" wins - overrides the property. So perhaps this order could be extended to multiple engine definitions. For example:

# default engine (optional, default is 0)
quarkus.vault.kv-secret-engine-order=0

# named engine (optional, default is 0)
quarkus.vault.kv-secret-engine."myengine".order=1

Engines are ordered by increasing order value. If there is a tie, engine name is used lexicographically with default engine's nonexistent name being ordered before all other named engines.

The engine ordered later wins over engines ordered before it like the config path defined later within an engine wins over config paths defined before it.

Even if no "order" is explicitly defined for engines, the ordering is deterministic.

WDYT?

ousgougou:
It is OK for me.
Thank you

plevart:
One more thing? Is it possible to define just named engines and somehow "disable" the default engine? I believe that the default engine would be used only in situations that require a single engine. When multiple engines are required, one would prefer to configure them in a uniform way.

vsevel:
I believe this is ready for review @kdubb 
@plevart and @ousgougou any chance you could try it out?

cescoffier (approved):
LGTM

PR #522, opened 2026-07-14
Add support for Vault Agent Auto-Auth

### LINKED ISSUE #162: Add support for Vault Agent's Auto-Auth feature
Vault Agent provides an Auto-Auth capability where the agent is responsible for managing and using the token but a proxy API is provided so that clients no longer need to submit an `X-Vault-Token` header with requests. quarkus-vault is currently configured to omit this header if the configured token is `null`, but I don't see a safe way of eliciting this behavior (please correct me if I've overlooked something.)

Would it be possible to update the `VaultRuntimeConfig#authentication()` signature to return an `Optional<VaultAuthenticationConfig>` and add a new configuration property that allows this to be `None`? My company's internal cloud service has stopped passing wrapped client tokens to cloud services in favor of an Auto-Auth-enabled Vault Agent sidecar (not my decision), so I don't currently see another authentication option that will work with quarkus-vault, as-is.

vsevel:
Fixes #162

## What

Introduces `quarkus.vault.authentication.none`, an explicit opt-in that makes the extension send requests to Vault without any `X-Vault-Token` header. This allows pointing `quarkus.vault.url` at a [Vault Agent](https://developer.hashicorp.com/vault/docs/agent-and-proxy/agent) listener with [Auto-Auth](https://developer.hashicorp.com/vault/docs/agent-and-proxy/autoauth) enabled: the agent authenticates on behalf of the application and injects the Vault token into proxied requests itself.

```properties
quarkus.vault.url=http://localhost:8100
quarkus.vault.authentication.none=true
```

## How

The low-level client already supports token-less requests when no token provider is set, so the change is confined to the extension layer:

- `VaultAuthenticationConfig`: new `none()` property (default `false`).
- `VaultAuthenticationType`: new `NONE` constant, returned by `VaultRuntimeConfig#getAuthenticationType()` when the flag is set.
- `VaultClientProducer`: when `none` is set, no token provider is installed; combining it with any other authentication settings fails with a descriptive error. A fully unconfigured authentication now also fails with a clear message (previously a `NullPointerException` from switching on a null authentication type).

An explicit flag was preferred over treating "no auth configured" as auto-auth mode, so that a typo'd authentication config keeps producing a startup error instead of silently sending unauthenticated requests.

## Testing

- `VaultAgentAutoAuthITCase` + `WiremockVaultAgent`: WireMock simulates the agent; the stub only matches requests where the `X-Vault-Token` header is **absent**, so any leaked token fails the test. The app reads a KV secret through the simulated agent.
- `VaultAuthenticationNoneConflictITCase`: asserts the error when `none` is combined with `client-token`.
- Full `integration-tests/vault` suite passes locally (the only errors are a pre-existing `VerifyError` in `VaultTestExtension.execLocalStack`, unrelated to this change).

## Docs

- New "Vault Agent Auto-Auth" section in `vault-auth.adoc` with a sample agent HCL configuration.
- Added the property to the generated config reference (`includes/quarkus-vault.adoc`) by hand, matching the generated format — note that `docs/pom.xml` still copies from the legacy `../target/asciidoc/generated/config/` path, which current Quarkus versions no longer produce, so local regeneration is a silent no-op.

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
@@ -691,6 +691,25 @@ h|[[quarkus-vault_quarkus-vault-authentication-authentication]]link:#quarkus-vau
 h|Type
 h|Default
 
+a| [[quarkus-vault_quarkus-vault-authentication-none]]`link:#quarkus-vault_quarkus-vault-authentication-none[quarkus.vault.authentication.none]`
+
+
+[.description]
+--
+Disable authentication: no client token will be sent to Vault. This is useful when the Vault url points to a Vault Agent with Auto-Auth enabled, which authenticates on behalf of the application and injects the Vault token into proxied requests. This property is exclusive with all other authentication settings.
+
+See link:https://developer.hashicorp.com/vault/docs/agent-and-proxy/autoauth[Vault Agent Auto-Auth]
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_VAULT_AUTHENTICATION_NONE+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_VAULT_AUTHENTICATION_NONE+++`
+endif::add-copy-button-to-env-var[]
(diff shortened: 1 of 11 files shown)


omatheusmesmo (approved):
LGTM!

PR #523, opened 2026-07-14
Add support for destroy in Key-Value engine

### LINKED ISSUE #77: Add support for `destroy` in Key-Value Engine
Hi team,

I am trying to completely remove a key-value pair from Vault, including removing the path. The `delete` method on the KVEngine only deactivates the secret, as far as I understood it. 

I suppose the `destroy` method would allow us to completely remove a secret from Vault. Is support for this method planned? I'd help out, but I would need some pointers on where to do this.

Kind regards!

vsevel:
Fixes #77

Exposes the kv v2 [destroy](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#destroy-secret-versions) operation on the high-level KV engine API (the low-level client already supported it via `destroySecretVersions`):

- `VaultKVSecretReactiveEngine.destroySecret(alias, path, versions)` + default-mount overload
- `VaultKVSecretEngine.destroySecret(path, versions)` (blocking wrapper)
- `KvV2` delegates to the client's `destroySecretVersions`; `KvV1` fails with `UnsupportedOperationException` since destroy only exists in kv v2 (v1 deletes are already permanent, and silently ignoring the versions argument would be misleading)

Tests (run against the Vault testcontainer):
- `VaultITCase#destroySecret` (kv v2): destroying a non-current version leaves the current version readable; destroying the current version makes reads return 404. Test policy gained entries for `secret/data/crud-destroy` and `secret/destroy/crud-destroy`.
- `VaultKv1ITCase#destroySecretV1NotSupported` (kv v1): asserts `UnsupportedOperationException`.

Note: destroy permanently removes version data but not the path metadata; fully erasing the path would be Vault's *delete metadata* operation, which could be a follow-up if desired.

### DIFF
--- a/integration-tests/vault-agroal/src/test/java/io/quarkus/vault/VaultKv1ITCase.java
+++ b/integration-tests/vault-agroal/src/test/java/io/quarkus/vault/VaultKv1ITCase.java
@@ -1,10 +1,13 @@
 package io.quarkus.vault;
 
+import java.util.List;
+
 import jakarta.inject.Inject;
 
 import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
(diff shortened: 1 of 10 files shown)


omatheusmesmo (approved):
LGTM!

PR #524, opened 2026-07-14
Support fetching the username from KV secrets in credentials provider

### LINKED ISSUE #102: credentials-provider is not loading username
When using credentials-provider which is default hashicorp vault, if I define get password from Vault it is working weel, but when I try to get datasource.username from Vault I am getting error

```
   %prod.quarkus.vault.url=http://localhost:8200
   %prod.quarkus.vault.authentication.userpass.username=<vault-username>
   %prod.quarkus.vault.authentication.userpass.password=<vault-password>
   %prod.quarkus.vault.credentials-provider.quarkus_banking.kv-path=myapps/quarkus/db-credentials
   %prod.quarkus.vault.secret-config-kv-path=myapps/quarkus/db-credentials

   %prod.quarkus.datasource.db-kind=postgresql
   %prod.quarkus.datasource.credentials-provider=quarkus_banking
```
 this one is not working, and if I add 

```
%prod.quarkus.datasource.username=<username>
```
It is working well.
This is my vault configurations:

```
Key                Value
---                -----
created_time       2023-01-03T18:42:13.992602027Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            8

====== Data ======
Key         Value
---         -----
password    <password>
user        <username>
```

I tried `username` keyword too.But the same result. Error decription:
```
 [io.qua.run.Application] (main) Failed to start application (with profile prod): org.postgresql.util.PSQLException: FATAL: password authentication failed for user "mysystemuser"
```

After digging into io.quarkus.vault.runtime.VaultCredentialsProvider  from quarkus-vault:2.1.0 lib 
I found this code 
```java
            if (config.kvPath.isPresent()) {
            String password = vaultKVSecretEngine.readSecret(config.kvPath.get()).get(config.kvKey);
            Map<String, String> result = new HashMap<>();
            result.put(PASSWORD_PROPERTY_NAME, password);
            return result;
        }
```
I think this is the route of the issue: value for `USER_PROPERTY_NAME` key will not be added to the final result when kvPath is present. But according to the [official guide](https://quarkus.io/guides/credentials-provider)  datasource username should be read from Vault if it is not exist in application.properties file, as guide says:
_Note that quarkus.datasource.username is the original agroal property, whereas the password property is not included because the value will come from the mydatabase credentials provider we just defined. An alternative is to define both username and password in Vault and drop the quarkus.datasource.username property from configuration. All consuming extensions do support the ability to fetch both the username and password from the provider, or just the password._
My quarkus version:  2.15.1
Java : amazon-corretto-17
Maven: 3.8.6

vsevel:
Fixes #102

Implements the approach proposed in https://github.com/quarkiverse/quarkus-vault/issues/102#issuecomment-1476620417 and https://github.com/quarkiverse/quarkus-vault/issues/102#issuecomment-1476636841:

- `quarkus.vault.credentials-provider."name".kv-key` is deprecated and its type changed to `Optional` (as discussed in the issue, this allows detecting whether it was explicitly set)
- new `kv-password-key` configuration, defaulting to `password`
- new `kv-username-key` configuration, defaulting to `username`

When the username key is present in the KV secret, `VaultCredentialsProvider` now returns it under the standard `user` property, so consuming extensions (e.g. Agroal) can fetch both the username and the password from Vault and `quarkus.datasource.username` can be dropped, as documented in the [Quarkus credentials provider guide](https://quarkus.io/guides/credentials-provider). Secrets without a username key behave exactly as before.

For backward compatibility, `kv-key`, when set, takes precedence over `kv-password-key`.

Testing:
- `VaultITCase.credentialsProvider` now covers three variants: default keys, explicit `kv-password-key`/`kv-username-key`, and the deprecated `kv-key`
- the shared `cred-provider.json` test fixture now contains a `username` entry, which makes the existing `vault-agroal` static datasource IT exercise username-from-Vault end-to-end against PostgreSQL
- `integration-tests/vault` and `integration-tests/vault-agroal` pass locally

Docs: the datasource guide documents the new keys and the username-in-Vault setup; the generated config reference was updated in sync with the javadoc.

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-vault.adoc
@@ -623,7 +623,7 @@ a| [[quarkus-vault_quarkus-vault-credentials-provider-credentials-provider-kv-pa
 
 [.description]
 --
-A path in vault kv store, where we will find the kv-key.
+A path in vault kv store, where we will find the kv-password-key and the kv-username-key.
 <p>
 One of `database-credentials-role` or `kv-path` needs to be defined. not both.
 <p>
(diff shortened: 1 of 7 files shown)


omatheusmesmo (commented):
Great job!

omatheusmesmo (approved):

PR #74, opened 2022-09-02
Add github authentication

r00ta:
STILL WIP - Early feedback is more than welcome: i.e. if this feature is a nice to have for the project I'm happy to move on with the integration tests. I've already tested it locally and it works fine. 

Relates to https://github.com/quarkiverse/quarkus-vault/issues/73 

This PR aims to provide GitHub as authentication mechanism for the vault.

PR #525, opened 2026-07-15
Add GitHub authentication support

### LINKED ISSUE #73: Is it possible to authenticate with a Github token?
Hi team, 

I'm trying to login with a Github token but I can't find the configuration to do it https://quarkiverse.github.io/quarkiverse-docs/quarkus-vault/dev/index.html#configuration-reference. 

Screenshot from 2022-09-01 15-00-23

Is it already available or it a missing feature? 

Thank you so much :pray:

vsevel:
Fixes https://github.com/quarkiverse/quarkus-vault/issues/73

Reimplements #74 on top of the current spec-driven client architecture, with integration tests.

## Client module

- New `auth/github.yaml` API spec covering Vault's GitHub auth API: `login`, `configure`/`readConfig` (organization, organization-id, base-url, token tuning) and team/user mapping CRUD + list, generating a `VaultAuthGithub` API reachable via `client.auth().github()`
- New `VaultGithubAuthOptions` and `VaultGithubTokenProvider` (mirroring the userpass pair), plus `github(token)` / `github(options)` methods on the `VaultClient` builder, with token caching/renewal and wrapped-token support

## Runtime extension

- New `GITHUB` authentication type, configured with:
  - `quarkus.vault.authentication.github.token`
  - `quarkus.vault.authentication.github.token-wrapping-token` (kv secret with a `token` key)
  - `quarkus.vault.authentication.github.auth-mount-path` (defaults to `github`)

## Tests

Vault's github auth method validates tokens against the GitHub API, but supports a custom `base_url`. Tests use a small JDK `HttpServer` GitHub API stub (`/user`, `/user/orgs`, `/user/teams`, `/orgs/{org}`) that Vault reaches from inside its container through testcontainers host access (`withAccessToHost(true)` + `host.testcontainers.internal`), allowing real end-to-end logins:

- `VaultAuthGithubTest` (client): config round-trip, organization-id auto-resolution through the GitHub API, team/user mappings, and end-to-end logins asserting metadata and mapped policies
- `VaultGithubITCase` / `VaultGithubWrapITCase` (integration-tests): boot Quarkus with github authentication (direct and wrapped token) and read a KV secret

## Docs

- New "GitHub Authentication" section in `vault-auth.adoc` and the new properties added to the config reference

### DIFF
--- a/client/src/test/java/io/quarkus/vault/client/VaultAuthGithubTest.java
+++ b/client/src/test/java/io/quarkus/vault/client/VaultAuthGithubTest.java
@@ -1,0 +1,223 @@
+package io.quarkus.vault.client;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.time.Duration;
+import java.util.concurrent.ExecutionException;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.vault.client.api.auth.github.VaultAuthGithubConfigureParams;
+import io.quarkus.vault.client.test.GithubMockServer;
+import io.quarkus.vault.client.test.Random;
+import io.quarkus.vault.client.test.VaultClientTest;
+import io.quarkus.vault.client.test.VaultClientTest.Mount;
+
+@VaultClientTest(auths = {
+        @Mount(type = "github", path = "github")
+})
+public class VaultAuthGithubTest {
+
+    private static final GithubMockServer githubMockServer = new GithubMockServer();
+
+    @BeforeAll
+    public static void startGithubMockServer() {
+        githubMockServer.start();
+    }
+
+    @AfterAll
+    public static void stopGithubMockServer() {
+        githubMockServer.close();
+    }
+
+    @Test
+    public void testConfig(VaultClient client) throws Exception {
+        var githubApi = client.auth().github();
+
+        githubApi.configure(new VaultAuthGithubConfigureParams()
+                .setOrganization(GithubMockServer.ORGANIZATION)
+                .setOrganizationId(GithubMockServer.ORGANIZATION_ID)
+                .setBaseUrl(githubMockServer.getBaseUrl())
+                .setTokenTtl(Duration.ofHours(1))
+                .setTokenMaxTtl(Duration.ofHours(2)))
+                .toCompletableFuture().get();
+
+        var config = githubApi.readConfig()
+                .toCompletableFuture().get();
+
+        assertThat(config.getOrganization())
+                .isEqualTo(GithubMockServer.ORGANIZATION);
+        assertThat(config.getOrganizationId())
+                .isEqualTo(GithubMockServer.ORGANIZATION_ID);
+        assertThat(config.getBaseUrl())
+                .isEqualTo(githubMockServer.getBaseUrl());
+        assertThat(config.getTokenTtl())
+                .isEqualTo(Duration.ofHours(1));
+        assertThat(config.getTokenMaxTtl())
+                .isEqualTo(Duration.ofHours(2));
+    }
+
+    @Test
+    public void testConfigResolvesOrganizationId(VaultClient client) throws Exception {
+        var githubApi = client.auth().github();
+
+        // organization_id is not provided, forcing Vault to resolve it against the (mock) GitHub API
+        githubApi.configure(new VaultAuthGithubConfigureParams()
+                .setOrganization(GithubMockServer.ORGANIZATION)
+                .setBaseUrl(githubMockServer.getBaseUrl()))
+                .toCompletableFuture().get();
+
+        var config = githubApi.readConfig()
+                .toCompletableFuture().get();
+
+        assertThat(config.getOrganizationId())
+                .isEqualTo(GithubMockServer.ORGANIZATION_ID);
+    }
+
+    @Test
+    public void testTeamMappings(VaultClient client, @Random String team) throws Exception {
+        var githubApi = client.auth().github();
+
+        var team1 = team.toLowerCase() + "1";
+        var team2 = team.toLowerCase() + "2";
+
+        githubApi.updateTeamMapping(team1, "default,dev")
+                .toCompletableFuture().get();
+        githubApi.updateTeamMapping(team2, "prod")
+                .toCompletableFuture().get();
+
+        var mapping = githubApi.readTeamMapping(team1)
+                .toCompletableFuture().get();
+
+        assertThat(mapping.getKey())
+                .isEqualTo(team1);
+        assertThat(mapping.getValue())
+                .isEqualTo("default,dev");
+
+        var teams = githubApi.listTeamMappings()
+                .toCompletableFuture().get();
+
+        assertThat(teams)
+                .contains(team1, team2);
+
+        githubApi.deleteTeamMapping(team1)
+                .toCompletableFuture().get();
+
+        var teams2 = githubApi.listTeamMappings()
+                .toCompletableFuture().get();
+
+        assertThat(teams2)
+                .doesNotContain(team1)
+                .contains(team2);
+    }
+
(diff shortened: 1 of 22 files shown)


vsevel:
hello @r00ta if this still of interest for you, can you review this PR and provide feedback? thanks.

client/src/test/java/io/quarkus/vault/client/test/GithubMockServer.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getBaseUrl() {
+        return "http://host.testcontainers.internal:" + server.getAddress().getPort() + "/";
+    }
+
+    @Override
+    public void close() {
+        server.stop(0);
+    }
+
+    private static void respond(HttpExchange exchange, String body) throws IOException {
```
omatheusmesmo: The mock answers 200 to any request, so it never checks the token Vault forwards. To see whether that hides real breakage, I patched `VaultGithubTokenProvider` to send a deliberately wrong token (`"MUTANT-" + token`) and the suite still passed 6/6, so `testLoginProcess` and `testClientLogin` can't fail for the one thing they exist to prove. `login()` is generated from `github.yaml` (`bodyFrom: [token]`), so a spec or generator change could break the token round trip with no test catching it. Suggest checking the token in `Authorization` on `/user`, `/user/orgs` and `/user/teams` and returning 401 otherwise, plus a negative test. Match the value but tolerate either scheme, since GitHub accepts both `Bearer <token>` and `token <token>`, and pinning the one Vault happens to send today would couple the test to its HTTP client.

```java
var header = exchange.getRequestHeaders().getFirst("Authorization");
var token = header == null ? null : header.replaceFirst("(?i)^(Bearer|token)\\s+", "");
if (!VALID_TOKEN.equals(token)) { /* 401 */ }
```

I tried this locally: the mutation above goes red, and the unmutated suite stays green. Only caveat is to leave `/orgs/{org}` unauthenticated, since Vault resolves the organization id there at config time, before any token exists. Checking it there fails with `unable to fetch the organization_id, you must manually set it in the config: GET /orgs/testorg: 401 Bad credentials`.

test-framework/src/main/java/io/quarkus/vault/test/GithubMockServer.java
```diff
@@ -0,0 +1,72 @@
...
+    public String getBaseUrl() {
+        return "http://host.testcontainers.internal:" + server.getAddress().getPort() + "/";
+    }
+
+    @Override
+    public void close() {
+        server.stop(0);
+    }
+
+    private static void respond(HttpExchange exchange, String body) throws IOException {
```
omatheusmesmo: Same gap as the mock under `client`: it answers 200 to any request, so it never checks that the token Vault forwards is the one that was configured. With `VaultGithubTokenProvider` patched to send a wrong token, `VaultGithubITCase` and `VaultGithubWrapITCase` both still pass. It matters more here, since these two cover the extension's real path, from `quarkus.vault.authentication.github.token` through `VaultClientProducer` to login, including response wrapping. Same fix: check the token on the `/user*` endpoints and return 401 otherwise.

client/src/test/java/io/quarkus/vault/client/test/GithubMockServer.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+        server.createContext("/user", exchange -> respond(exchange, USER));
+        server.createContext("/user/orgs", exchange -> respond(exchange, USER_ORGS));
+        server.createContext("/user/teams", exchange -> respond(exchange, USER_TEAMS));
+        server.createContext("/orgs/" + ORGANIZATION, exchange -> respond(exchange, ORG));
+    }
+
+    public void start() {
+        server.start();
+        org.testcontainers.Testcontainers.exposeHostPorts(server.getAddress().getPort());
```
omatheusmesmo: Nit: inline FQCN instead of an import, `org.testcontainers.Testcontainers.exposeHostPorts(...)`.

test-framework/src/main/java/io/quarkus/vault/test/GithubMockServer.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        server.createContext("/user", exchange -> respond(exchange, USER));
+        server.createContext("/user/orgs", exchange -> respond(exchange, USER_ORGS));
+        server.createContext("/user/teams", exchange -> respond(exchange, USER_TEAMS));
+        server.createContext("/orgs/" + ORGANIZATION, exchange -> respond(exchange, ORG));
+    }
+
+    public void start() {
+        server.start();
+        org.testcontainers.Testcontainers.exposeHostPorts(server.getAddress().getPort());
```
omatheusmesmo: Nit: inline FQCN instead of an import, `org.testcontainers.Testcontainers.exposeHostPorts(...)`.

client/src/test/java/io/quarkus/vault/client/test/GithubMockServer.java
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.charset.StandardCharsets;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpServer;
+
+/**
+ * A minimal GitHub API stub implementing the endpoints used by Vault's github auth method:
+ * user lookup, user organizations, user teams and organization lookup. Vault is pointed at it
+ * via the github auth method's {@code base_url} configuration property, using the
+ * {@code host.docker.internal} hostname from inside the Vault container.
```
omatheusmesmo: Nit: javadoc says `host.docker.internal`, code uses `host.testcontainers.internal` (line 58). The test-framework copy is correct.

client/src/test/java/io/quarkus/vault/client/VaultAuthGithubTest.java
```diff
@@ -0,0 +1,223 @@
...
+        assertThat(config.getBaseUrl())
+                .isEqualTo(githubMockServer.getBaseUrl());
+        assertThat(config.getTokenTtl())
+                .isEqualTo(Duration.ofHours(1));
+        assertThat(config.getTokenMaxTtl())
+                .isEqualTo(Duration.ofHours(2));
+    }
+
+    @Test
+    public void testConfigResolvesOrganizationId(VaultClient client) throws Exception {
```
omatheusmesmo: Nit: the comment on line 67 says it forces Vault to resolve the org id against the mock, but `/orgs/testorg` is never called in a full suite run, only when the test runs in isolation. Vault merges into the stored config and resolves only `if c.OrganizationID == 0` (`builtin/credential/github/path_config.go:116`, v1.15.4), and by the time this runs another test in the class has already written `organization_id=456` to the shared `github` mount. Giving the test its own mount does make the resolution fire. Even then the assertion is weak: sending `organization_id=0` explicitly also resolves to 456, so `== 456` can't tell an omitted field from a zero one and only catches a wrong non zero value. Removing it is a fair option too.

omatheusmesmo (commented):
Good overall! I verified the production code by running it: the token is transmitted correctly. A few test adjustments below.

client/src/test/java/io/quarkus/vault/client/test/GithubMockServer.java
vsevel: that is correct. the initial implementation used `host.docker.internal`. this was later replaced, but the doc was not updated.

client/src/test/java/io/quarkus/vault/client/VaultAuthGithubTest.java
vsevel: good catch. the fix suggestion was to to create a dedicated mount or remove the test as you suggested, which I did.

vsevel:
@omatheusmesmo thanks for your accurate review. all points have been taken into account. any more feedback on that one?

r00ta:
> hello @r00ta if this still of interest for you, can you review this PR and provide feedback? thanks.

Hey, I no longer work for RedHat, so I'll leave it to the current maintainers

omatheusmesmo (approved):
LGTM @vsevel , I would advice only to squash the commits before merging it.

PR #526, opened 2026-07-15
Reproducer for #328: renew-grace-period ignored with approle authentication

### LINKED ISSUE #328: Question: Why is default value of token refresh 30 seconds?

vsevel:
Reproducer for #328.

## Root cause

`VaultClientProducer.configureAuthentication` propagates the configured `quarkus.vault.renew-grace-period` to the client for KUBERNETES and USERPASS via `.caching(config.renewGracePeriod())`, but never for **APPROLE**, which therefore keeps the hard-coded 30s default (`VaultCachingTokenProvider.DEFAULT_RENEW_GRACE_PERIOD`).

With approle, the login token is only extended when a Vault access happens within 30s of its expiry, whatever `quarkus.vault.renew-grace-period` says. If no call lands in that window the token expires, and Vault revokes the leases attached to it — notably dynamic database credentials — after which DB accesses fail even though the credentials lease was still advertised as valid. This matches the reporter's symptom ("The server principal \"v-approle-...\" is not able to access the database under the current security context"), and their workaround (patching `DEFAULT_RENEW_GRACE_PERIOD`).

## The new `integration-tests/vault-renew` module

Two scenarios, each run once with approle (**fails** on current main) and once with userpass (passes, as the control — the auth method is the only difference):

- `*RenewGracePeriodITCase` — pinpoint scenario against an in-process Vault HTTP stub that records login/renew-self calls (60s token TTL, `renew-grace-period=40s`): a Vault access made ~35s before token expiry — inside the configured grace period but outside the hard-coded 30s — must extend the token. With approle no renew-self is sent.
- `*DynamicCredentialsITCase` — end-to-end scenario with a real Vault and PostgreSQL (testcontainers), the datasource configured with Vault dynamic credentials (90s token TTL, `renew-grace-period=60s`, 10m credentials lease). With approle the token isn't extended, expires, Vault revokes the credentials lease with it and runs the revocation statements — the next DB access gets `ERROR: permission denied for table demo`.

See `integration-tests/vault-renew/README.md` for the full analysis and timelines.

## Observed results on current main

```
[ERROR]   AppRoleRenewGracePeriodITCase ... expected: <1> but was: <0>  (no renew-self sent)
[ERROR]   AppRoleDynamicCredentialsITCase ... Response: DB access failed: org.postgresql.util.PSQLException:
          ERROR: permission denied for table demo ==> expected: <200> but was: <500>
```

Both userpass variants pass.

## Fix

The one-line fix (verified locally: all four tests pass with it) would be, in the APPROLE branch of `VaultClientProducer.configureAuthentication`:

```java
appRoleOptions.caching(config.renewGracePeriod());
```

It is intentionally not included in this PR so the tests demonstrate the issue; happy to add it (with the tests then acting as the regression tests) if preferred.

Fixes #328 (once the fix is applied)

### DIFF
--- a/docs/modules/ROOT/pages/includes/attributes.adoc
+++ b/docs/modules/ROOT/pages/includes/attributes.adoc
@@ -1,4 +1,4 @@
-:quarkus-version: 3.36.2
+:quarkus-version: 3.37.2
 :quarkus-vault-version: 4.8.0
 :maven-version: 3.8.1+
 
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -13,5 +13,6 @@
         <module>vault</module>
         <module>vault-agroal</module>
         <module>vault-app</module>
+        <module>vault-renew</module>
     </modules>
 </project>
--- a/integration-tests/vault-renew/README.md
+++ b/integration-tests/vault-renew/README.md
@@ -1,0 +1,113 @@
+# Reproducer for issue #328 — `quarkus.vault.renew-grace-period` ignored with AppRole authentication
+
+https://github.com/quarkiverse/quarkus-vault/issues/328
+
+## The problem reported by @Visserr2
+
+With approle authentication, the login token is only extended (renew-self) when a Vault access
+happens within **30 seconds** of the token's expiry, even when `quarkus.vault.renew-grace-period`
+is set to a larger value (the reporter used 1m, and 35m for long-running jobs). If no Vault call
+lands inside that 30s window, the token expires; leases attached to it (e.g. dynamic database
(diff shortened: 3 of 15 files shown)


vsevel:
closing as we now have a PR for that issue https://github.com/quarkiverse/quarkus-vault/pull/527

PR #527, opened 2026-07-15
Honor renew-grace-period with approle authentication

### LINKED ISSUE #328: Question: Why is default value of token refresh 30 seconds?

vsevel:
Fixes https://github.com/quarkiverse/quarkus-vault/issues/328

## Root cause

`VaultClientProducer.configureAuthentication` propagates the configured `quarkus.vault.renew-grace-period` to the client for KUBERNETES and USERPASS via `.caching(config.renewGracePeriod())`, but never did for **APPROLE**, which therefore fell back to the hard-coded 30s default (`VaultCachingTokenProvider.DEFAULT_RENEW_GRACE_PERIOD`).

As a consequence, with approle the login token was only extended when a Vault access happened within 30s of its expiry, whatever the configured grace period. If no call landed in that window the token expired, and Vault revoked the leases attached to it — notably dynamic database credentials — after which DB accesses failed even though the credentials lease was still advertised as valid ("The server principal \"v-approle-...\" is not able to access the database under the current security context").

The issue is demonstrated by the reproducer in #526: on current `main`, its two approle integration tests fail in CI (no renew-self inside the configured grace period; `permission denied for table demo` after Vault revokes the dynamic DB credentials lease), while the identical userpass scenarios pass. With this one-line fix, all four reproducer tests pass.

## The fix

```java
appRoleOptions.caching(config.renewGracePeriod());
```

in the APPROLE branch, matching what the KUBERNETES and USERPASS branches already do.

## The test

`AppRoleRenewGracePeriodTest` is a fast `@QuarkusTest` in the deployment module (~3s, no containers): a minimal in-process Vault stub issues approle tokens with a 60s TTL, and `quarkus.vault.renew-grace-period` is set to 59s. The first KV read logs in; a second read 2s later finds the token expiring inside the configured grace period and must extend it (renew-self). Without the fix no renewal happens, since the token is not within 30s of expiry: `expected: <logins=1;renewals=1> but was: <logins=1;renewals=0>`.

### DIFF
--- a/deployment/src/test/java/io/quarkus/vault/AppRoleRenewGracePeriodTest.java
+++ b/deployment/src/test/java/io/quarkus/vault/AppRoleRenewGracePeriodTest.java
@@ -10,18 +10,19 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import jakarta.inject.Inject;
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Test;
 
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpServer;
 
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
 import io.quarkus.test.junit.QuarkusTest;
 import io.quarkus.test.junit.QuarkusTestProfile;
 import io.quarkus.test.junit.TestProfile;
@@ -37,9 +38,9 @@ import io.quarkus.test.junit.TestProfile;
  * client's hard-coded 30s default, no renewal happens (the token still has ~58s to live) and this
  * test fails.
  * <p>
- * Note: the test class and the test profile run in different classloaders, each with its own copy
- * of the stub's statics, so the test reads the stub's counters over HTTP instead of accessing them
- * directly.
+ * Note: the test class and the test resource managing the stub run in different classloaders, each
+ * with its own copy of the stub's statics, so the test reads the stub's counters over HTTP instead
+ * of accessing them directly.
  */
 @QuarkusTest
 @TestProfile(AppRoleRenewGracePeriodTest.Profile.class)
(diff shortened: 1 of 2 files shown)


deployment/src/test/java/io/quarkus/vault/AppRoleRenewGracePeriodTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private String counters() throws IOException, InterruptedException {
+        return HttpClient.newHttpClient()
+                .send(HttpRequest.newBuilder(URI.create(vaultUrl + "/stub/counters")).build(),
+                        HttpResponse.BodyHandlers.ofString())
+                .body();
+    }
+
+    @AfterAll
+    public static void stopStub() {
```
omatheusmesmo: Optional: because the `Profile` runs in more than one classloader, `start()` binds a stub server per classloader but `@AfterAll stopStub()` closes only the runtime-classloader copy, so one `HttpServer` leaks its thread/port for the rest of the JVM. Harmless here. A daemon executor on the stub makes a missed `stop()` leak nothing live:

```java
server.setExecutor(Executors.newCachedThreadPool(r -> {
    var t = new Thread(r, "vault-stub");
    t.setDaemon(true);
    return t;
}));
```

omatheusmesmo (approved):
LGTM! One optional comment.

deployment/src/test/java/io/quarkus/vault/AppRoleRenewGracePeriodTest.java
vsevel: Good catch, thanks. I went one step further than the daemon executor: it would only cover the handler threads, while the `HttpServer`'s internal `HTTP-Dispatcher` thread is never marked daemon, so a missed `stop()` would still leave a live non-daemon thread. The root cause was the unpaired start/stop across classloaders, so in b6d19ad the stub server lifecycle moved into a `QuarkusTestResourceLifecycleManager` registered via `Profile.testResources()`: the framework guarantees `stop()` is called on the same instance it started, the server is an instance field instead of a per-classloader static, and the `@AfterAll` is gone. No server copy can outlive the test anymore.

omatheusmesmo (approved):
LGTM! I would only squash the commits, but it's up to you.

vsevel:
https://github.com/all-contributors add @omatheusmesmo  for code

vsevel:
@all-contributors add @omatheusmesmo for code

PR #537, opened 2026-07-21
Fix Invalid PEM Certificate when ca-cert holds multiple certificates

### LINKED ISSUE #347: VaultException: Invalid PEM Certificate, when a pem file contains more than one certificate
Hi,

**quarkus** 3.15
**quarkus-vault** 4.1.0  

If I set quarkus.vault.tls.ca-cert to a pem file and the file contains more than one certificate, then I got VaultException: Invalid PEM Certificate.

vsevel:
## Problem

When `quarkus.vault.tls.ca-cert` points to a PEM file containing more than one certificate (e.g. a CA chain of a root plus an intermediate), the JDK-based Vault HTTP client fails with `VaultException: Invalid PEM Certificate`.

Fixes #347

## Root cause

`JDKClientFactory.buildSslContextFromPem()` parsed the ca-cert file with the single-certificate parser `X509Parsing.parsePEMCertificate()`, which validates the whole file against a one-block regexp via `Matcher.matches()` and therefore rejects any file holding more than one `BEGIN/END CERTIFICATE` block.

## Fix

- Parse the ca-cert file with `X509Parsing.parsePEMCertificates()` (the `find()`-based variant) and register every certificate in the trust store under a distinct alias, so a full CA chain is trusted.
- Guard against an empty file with a clear error message.

The single-certificate `parsePEMCertificate()` is left unchanged, since it is also used by `CertificateData` in the PKI secret engine where exactly one certificate is expected.

The Vert.x-based client (`MutinyVertxClientFactory`) already handled multi-certificate files via `PemTrustOptions.addCertPath`, so it is unchanged.

## Tests

- `X509ParsingTest` — documents that `parsePEMCertificate` accepts only one block while `parsePEMCertificates` handles a chain.
- `JDKClientFactoryTest` — regression test writing a two-certificate PEM to a temp file and asserting `buildSslContextFromPem` now builds a valid `SSLContext` (previously threw `VaultException`).

### DIFF
--- a/runtime/src/main/java/io/quarkus/vault/runtime/client/JDKClientFactory.java
+++ b/runtime/src/main/java/io/quarkus/vault/runtime/client/JDKClientFactory.java
@@ -56,18 +56,23 @@ public class JDKClientFactory {
         }
     }
 
-    private static SSLContext buildSslContextFromPem(String file) throws VaultException {
+    static SSLContext buildSslContextFromPem(String file) throws VaultException {
         try {
 
-            final var certificate = X509Parsing.parsePEMCertificate(Files.readString(Paths.get(file), UTF_8));
+            final var certificates = X509Parsing.parsePEMCertificates(Files.readString(Paths.get(file), UTF_8));
+            if (certificates.isEmpty()) {
+                throw new CertificateException("No certificate found in " + file);
+            }
 
             final TrustManagerFactory trustManagerFactory = TrustManagerFactory
                     .getInstance(TrustManagerFactory.getDefaultAlgorithm());
 
-            // Build a truststore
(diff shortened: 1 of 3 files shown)


tafy2392:
👍  Tested the fix from PR by building the branch locally and deploying as a container. With quarkus.vault.tls.ca-cert pointing to a PEM file containing a certificate chain (root + intermediate), the VaultException: Invalid PEM Certificate no longer occurs. The fix works as expected. All other aspects of reading secrets are working fine

vsevel:
@tafy2392 thanks for your feedback.
@omatheusmesmo any chance you could give us a review?

omatheusmesmo (approved):

